PR #168, opened 2023-08-25
Refactor get connection names

lumoslnt:
Refactor get connection names.

In our current implementation, we generate tools twice: once when creating a Flow using `Flow.from_yaml`, and again when creating a FlowExecutor and resolving the tool. To improve this, we have removed the former step. However, this change breaks the logic of retrieving connection names, as we still need to find the connection type input in the tool. To address this issue, we have refactored the logic for retrieving connection names that we can identify the connection type input in the tool by simply loading the tool without complicated resolving logics.


thy09 (approved):

src/promptflow/promptflow/_core/tools_manager.py
```diff
@@ -249,6 +251,54 @@ def _load_custom_tool(tool: Tool, node_name: str) -> Callable:
...
 
+class ToolLoader:
+    def __init__(self, working_dir: str, package_tool_keys: Optional[List[str]] = None) -> None:
+        self._working_dir = working_dir
+        self._package_tools = collect_package_tools(package_tool_keys) if package_tool_keys else {}
+
+    def load_tool_for_node(self, node: Node) -> Tool:
+        if node.source is None:
+            raise UserErrorException(f"Node {node.name} does not have source defined.")
+        if node.type is ToolType.PYTHON:
```
zzn2, line 262: We should use `==` to replace `is` here.

See https://stackoverflow.com/a/133024 for details.

src/promptflow/tests/executor/unittests/_core/test_tools_manager.py
```diff
@@ -0,0 +1,88 @@
...
+from promptflow._core._errors import PackageToolNotFoundError
+from promptflow._core.tools_manager import ToolLoader
+from promptflow.contracts.flow import Node, ToolSource, ToolSourceType
+from promptflow.contracts.tool import Tool, ToolType
+from promptflow.exceptions import UserErrorException
+
+
+@pytest.mark.unittest
+class TestToolLoader:
+    def test_load_tool_for_node_with_invalid_node(self):
```
zzn2, line 15: Multi cases are tested in this method. Consider using pytest.parametrize.

src/promptflow/promptflow/_core/tools_manager.py
```diff
@@ -249,6 +251,54 @@ def _load_custom_tool(tool: Tool, node_name: str) -> Callable:
...
+            if node.source.type == ToolSourceType.Package:
+                return self.load_tool_for_package_node(node)
+            elif node.source.type == ToolSourceType.Code:
+                _, tool = self.load_tool_for_script_node(node)
```
zzn2, line 266: `load_tool_for_package_node` and `load_tool_for_script_node` returns different data types (one returns tool only and the other return both callable and tool). Do we have a plan to make them return the same data type?

src/promptflow/tests/executor/unittests/_core/test_tools_manager.py
lumoslnt: I have another pr to add unit tests for tool-related code, will fix in there.

src/promptflow/promptflow/_core/tools_manager.py
zzn2: Sorry, I was wrong. Enums can be compared using `is`.

https://stackoverflow.com/a/42866806

src/promptflow/promptflow/_core/tools_manager.py
lumoslnt: I am planning to refactor the code for generating tool meta, loading and resolving tools. I will try to fix this issue by the way.

src/promptflow/promptflow/_core/tools_manager.py
lumoslnt: Fixed.

zzn2 (approved):

thy09 (approved):

PR #241, opened 2023-08-30
[Flow test] Remove private function for streaming output

lalala123123:
# Description
1. Remove private function of executor in flow test
2. Show user code output in the terminal
   
image

3. Add streaming gif to doc

- [ ] **The pull request does not introduce [breaking changes]**

### Testing Guidelines


wangchao1230 (approved):

src/promptflow/promptflow/_sdk/operations/_test_submitter.py
```diff
@@ -216,57 +204,57 @@ def change_logger_level(level):
...
-            sys.stdout.write = write
+        def show_node_log_and_output(node_run_infos, show_node_output):
+            """Show stdout and output of nodes."""
+            for node_name, node_result in node_run_infos.items():
+                # Prefix of node stdout is "%Y-%m-%dT%H:%M:%S%z"
+                pattern = r"\[\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\+\d{4}\] "
+                node_logs = re.sub(pattern, "", node_result.logs["stdout"])
+                if node_logs:
+                    for log in node_logs.rstrip("\n").split("\n"):
+                        print(f"{Fore.LIGHTBLUE_EX}[{node_name}]:", end=" ")
```
crazygao, line 215: This may behave differently in different shells, please double confirm.

crazygao: especially output to bash/powershell/notebook.

lalala123123: https://github.com/tartley/colorama
colorma is used to cross-platform colored terminal text in Python

D-W-: @lalala123123 have we tested in those shells? Seems they are supported according to colorama's doc
image

src/promptflow/promptflow/_sdk/operations/_test_submitter.py
```diff
@@ -216,57 +204,57 @@ def change_logger_level(level):
...
-                write(*args, **kwargs)
-
-            sys.stdout.write = prefix_output
-            yield
-            sys.stdout.write = write
+        def show_node_log_and_output(node_run_infos, show_node_output):
+            """Show stdout and output of nodes."""
+            for node_name, node_result in node_run_infos.items():
+                # Prefix of node stdout is "%Y-%m-%dT%H:%M:%S%z"
+                pattern = r"\[\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\+\d{4}\] "
```
zzn2, line 211: It is not a good idea to depend on the output format here. If the format changes, this part will break.

src/promptflow/promptflow/_sdk/operations/_test_submitter.py
```diff
@@ -216,57 +204,57 @@ def change_logger_level(level):
...
-            else:
-                return generator
+                print(output)
+
+        def get_result_output(output):
+            if isinstance(output, GeneratorType):
+                if output in generator_record:
+                    output = iter(generator_record[output].items)
+                else:
+                    proxy = output.gi_frame.f_locals["proxy"]
```
zzn2, line 241: We may avoid using the internal field `gi_frame` of GeneratorType.

src/promptflow/promptflow/_sdk/operations/_test_submitter.py
lalala123123: Yes, it is not a good idea to depend on the format of stdout. But the stdout of user code can only be obtained through the line result.
If the stdout format changed, it will not break print the output here. It will directly print the stdout

src/promptflow/promptflow/_sdk/operations/_test_submitter.py
lalala123123: Not found other better way to get the input value of generator.
And gi_frame is exposed by generator type.
https://docs.python.org/3.8/library/inspect.html

docs/how-to-guides/init-and-test-a-flow.md
```diff
@@ -243,6 +243,38 @@ If a flow contains chat inputs or chat outputs in the flow interface, there will
...
+
+```python
+from promptflow import PFClient
+
+pf_client = PFClient()
+
+# Test flow
+inputs = {"<flow_input_name>": "<flow_input_value>"}  # The inputs of the flow.
+flow_result = pf_client.test(flow="<flow_folder_path>", inputs=inputs)
+for item in flow_result["<LLM node output>"]:
```
0mza987, line 270: what does this `<LLM node output>` should be? Isn't is the node name?

0mza987 (approved):

wangchao1230 (approved):

PR #288, opened 2023-09-04
[Executor] Calculating openai metrics under streaming mode

lumoslnt:
When using the OpenAI API in streaming mode, the response from OpenAI does not include metrics such as prompt tokens, completion tokens, and total tokens, so we need to calculate these metrics based on the input and output data.


src/promptflow/promptflow/_utils/openai_metrics_calculator.py
```diff
@@ -0,0 +1,159 @@
...
+            return self._get_openai_metrics_for_completion_api(api_call)
+        else:
+            self._log_warning(f"Cannot calculate metrics for api: {name}.")
+            return {}
+
+    def _get_openai_metrics_for_chat_api(self, api_call):
+        inputs = api_call.get("inputs")
+        output = api_call.get("output")
+        metrics = {}
+        enc, tokens_per_message, tokens_per_name = self._get_encoding_for_chat_api(inputs["engine"])
```
thy09, line 51: We should use api_type to decide whether use engine or model

lumoslnt: Updated.

thy09 (approved):

zzn2 (approved):

zzn2 (approved):

thy09 (approved):

PR #336, opened 2023-09-07
Remove None value before returning connection names

lumoslnt:
Remove None value when get connection names.


src/promptflow/promptflow/contracts/flow.py
```diff
@@ -515,7 +515,7 @@ def get_connection_names(self):
             tool = self.get_tool(node.tool) or self._tool_loader.load_tool_for_node(node)
             if tool:
                 connection_names.update(self._get_connection_name_from_tool(tool, node).values())
-        return connection_names
+        return set({item for item in connection_names if item})
```
thy09, line 518: test case?

thy09: Let's add it in the next PR

thy09 (approved):

huaiyan (approved):

PR #397, opened 2023-09-14
Fix tools_continuous_delivery when pr merge trigger.

liuyuhang13:
# Description
tools_continous_delivery will fail when pr merge trigger, because the trigger event is 'push' not 'pull_request'.
The called workflow still judge 'pull request'
image

Update to 'push' to fix this issue.
Update Version to workflow run number as it's not easy to get pull_request.number
Remove "Release" type, remove approval task, only reserve "test" for manual release.


chenslucky (dismissed):

.github/workflows/tools_continuous_delivery.yml
```diff
@@ -34,28 +26,12 @@ on:
         type: string
```
16oeahr, line 26: Do we still need dispatch if we only want continuous delivery?

liuyuhang13: removed

16oeahr (approved):

chjinche (approved):

PR #524, opened 2023-09-15
New tutorial on prompt tunning and evaluation

jiaochenlu:
# Description

- [x] **The pull request does not introduce [breaking changes]**

### Testing Guidelines


jiaochenlu:
@ttthree  @lisagreenview 
Please feel free to review this pull request, which includes the following changes:
1. Added a banner image (which has already been used as the GitHub social preview picture) to the Readme.
2. Contributed a new tutorial to highlight the value of batch testing and evaluation in PromptFlow, based on the simplest chat case. Additionally, we have created a video demo for better understanding.

scripts/readme/readme.py
```diff
@@ -275,7 +275,8 @@ def write_readme(workflow_telemetry, readme_telemetry):
     input_glob_readme = [
         "examples/flows/**/README.md",
         "examples/connections/**/README.md",
-        "examples/tutorials/**/chat*.md",
+        "examples/tutorials/e2e-development/*.md",
```
eniac871, line 278: Why do you need to change this?

examples/flows/evaluation/eval-chat-math/line_process.py
```diff
@@ -0,0 +1,43 @@
+from promptflow import tool
+
+
+@tool
+def line_process(groundtruth: str, prediction: str) -> int:
```
eniac871, line 5: please add some comments about the cases you may meet of groundtruth and prediction

examples/flows/chat/chat-math-variant/README.md
```diff
@@ -0,0 +1,50 @@
+# Test your prompt variants for chat with math
+This is a prompt tuning case with 3 prompt variants for math question answering.
```
eniac871, line 2: Can you elaborate more in somewhere to let users know how to do prompt tunning for this case? Otherwise, I don't see its big different compare with basic-char or chat-with-wikipedia,  it's still a standard introduction with another chat flow. In my understanding this case plus the evaluation flow is to show how to tune a chat flow

eniac871 (changes requested):

scripts/readme/readme.py
ChenJieting: It is because that we aim to show the index in examples/readme.md and tried to run the script examples/readme.py, but encountered some issues. Consequently, we reached out to @crazygao  for help.

crazygao: Not all tutorials are included in readme generation. I made changes to take this README into account

examples/flows/chat/chat-math-variant/README.md
jiaochenlu: This example aims to emphasize the importance of batch testing and evaluation in PromptFlow, using the most basic chat scenario as a foundation. 

Our aim is to enable users to quickly grasp the advantages of prompt tuning and experimentation with pf.

We provided [tutorial.md](https://github.com/microsoft/promptflow/pull/524/files?short_path=4071601#diff-40716014562432ce15ecf45dcc28e5b1909317cc631e7d2174e29ffc51252127),  introducing the prompt engineering and tunning approcah (Chain of thought). Here, we offer an in-depth explanation and introduction to the case. Moreover, we have produced a video demonstration that showcases the development of the flow using the VSC extension.

examples/flows/evaluation/eval-chat-math/line_process.py
jiaochenlu: added main function

ChenJieting (approved):

examples/flows/evaluation/eval-chat-math/line_process.py
eniac871: I mean some examples of the input, like for groundtruth what's the possible format you can handle? pure number? and any other format? I ask this because I see a large part of the logic is handle float parse exceptions, so I guess you meet some cases frequently which is not pure numbers. Put some samples values in the comments will be helpful for user understand the function

scripts/readme/readme.py
eniac871: then why do you delete the chat*.md?

examples/flows/chat/chat-math-variant/README.md
eniac871: Thanks for the explain, I mean if user see the example directly , maybe they don't know the benefit of it. What about add a refer link?

eniac871 (changes requested):

examples/flows/chat/chat-math-variant/README.md
jiaochenlu: Good point, it would be better to add a link to the tutorial to the process readme @ChenJieting let's do it

examples/flows/evaluation/eval-chat-math/line_process.py
jiaochenlu: Got it, we'll add some comments in the code line

examples/flows/chat/chat-math-variant/README.md
ChenJieting: I have added more explanation in chat-math-variant/README.md file.

examples/flows/evaluation/eval-chat-math/line_process.py
ChenJieting: Have added some comments in the code.

eniac871: hi Chenlu and Jieting, thanks for the follow up, the code itself is clear and don't need to comment. I mean you just need to put some examples in comments before the whole logic. For example 
# process prediction 
# will try parse the prediction and groundtruth to float. support parse ones like below: 0.2, 4/10, 1.0/3, 6E-1, 6E-1/0.6 ....

examples/flows/evaluation/eval-chat-math/line_process.py
```diff
@@ -0,0 +1,67 @@
...
+                    # for the wrong format with non-numeric numerator or zero denominator
+                    processed_result = -1
+            else:
+                # for the wrong format with multiple '/'
+                processed_result = -1 
+        else:
+            # for the wrong format without '/'
+            processed_result = -1
+
+    # process groundtruth
```
eniac871, line 29: Seems the process groundtruth and prediction logic are same. You can use a function to do this and just call it with different parameters

eniac871 (changes requested):

examples/flows/evaluation/eval-chat-math/line_process.py
ChenJieting: Have used a function to solve the same logic and refine the comment. Thanks for your suggestions.

eniac871: the string_to_number function actually don't need to pass into processed_result parameter, and you don't actually need a list : If the parsing is invalid, simple return a None should be fine. If it's valid, return the float value. 

def string_to_number(raw_string: str) -> float:
    ''' Try to parse the prediction string and groundtruth string to float number. 
    Support parse int, float, fraction and recognize non-numeric string with wrong format.
    For example: '3/5', '6/10', '0.5', 'the answer is \box{2/3}', '4/7//8'
    '''
    processed_result = 0
    float_number = 0.0
    try:
        float_number = float(raw_string) 
    except Exception:
        if '/' in raw_string: 
            split_list = raw_string.split('/') 
            if len(split_list) == 2: 
                numerator, denominator = split_list
                try:
                    float_number = float(numerator) / float(denominator) 
                except Exception:
                    return None
            else:
                return None
        else:
            return None
    return float_number

pred_float  = string_to_number('2/3')
''' Early stop'''
if (pred_float is None)
    return -1
    
gt_float = string_to_number('4/7//8')
if (gt_float is None)
    return -1
    
''' both pred_float and  gt_float are valid'''
if round(pred_float, 10) == round(gt_float, 10): 
            processed_result = 1 
        else:
            processed_result = -1

examples/flows/evaluation/eval-chat-math/line_process.py
```diff
@@ -0,0 +1,55 @@
+from promptflow import tool
+
+def string_to_number(raw_string: str, processed_result: int) -> list:
+    ''' Try to parse the prediction string and groundtruth string to float number. 
+    Support parse int, float, fraction and recognize non-numeric string with wrong format.
+    For example: '3/5', '6/10', '0.5', 'the answer is \box{2/3}', '4/7//8'
```
eniac871, line 6: these cases should not be able to parse one? If so, better to mention to clarify these're the wrong format case

eniac871 (changes requested):

scripts/readme/readme.py
eniac871: what's the update of this?

eniac871 (changes requested):

examples/flows/evaluation/eval-chat-math/line_process.py
ChenJieting: Thanks, that will be better. Have updated.

scripts/readme/readme.py
crazygao: they are the same, chat-with-pdf is also contained inside e2e-development folder.

why this is strict, because we actually don't want user to add new readme under examples/tutorials folder, this folder is originally meant for notebooks.

examples/flows/evaluation/eval-chat-math/line_process.py
ChenJieting: Have modified the code according to your suggestions. Thanks!

.github/workflows/samples_flows_chat_chat_math_variant.yml
```diff
@@ -0,0 +1,97 @@
...
+        uses: actions/checkout@v3
+      - name: Setup Python 3.9 environment
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.9"
+      - name: Generate config.json
+<<<<<<< HEAD
+        run: echo ${{ secrets.TEST_WORKSPACE_CONFIG_JSON }} > ${{ github.workspace }}/examples/config.json
+=======
+        run: echo '${{ secrets.TEST_WORKSPACE_CONFIG_JSON_CANARY }}' > ${{ github.workspace }}/examples/config.json
```
eniac871, line 29: How does this works?

eniac871 (changes requested):

.github/workflows/samples_flows_chat_chat_math_variant.yml
ChenJieting: @crazygao Could you please take a look on this?

ChenJieting: Have resolved the conflicts.

crazygao (approved):

eniac871 (approved):

PR #560, opened 2023-09-19
[Executor] Support ToolProvider for script tool

lumoslnt:
# Description

In this pull request, we have added support for `ToolProvider` to the script tool. This allows users to initialize their script tool before running it. With this update, users can define a script tool using a class method and initialize the tool in the *\_\_init\_\_()* method of the class.

This is a sample script tool:

```python
from promptflow import ToolProvider, tool


class ScriptToolWithInit(ToolProvider):
    def __init__(self, init_input: str):
        # This will only be called once when init the flow
        super().__init__()
        self.init_input = init_input  

    @tool
    def call(self, input: str):
        # This will be called multiple times
        return str.join(" ", [self.init_input, input])
```

- [x] **The pull request does not introduce [breaking changes]**

### Testing Guidelines


src/promptflow/promptflow/_core/tool_meta_generator.py
```diff
@@ -184,6 +194,21 @@ def collect_tool_function_in_module(m):
...
+    num_tools = len(functions) + len(methods)
+    if num_tools == 0:
+        raise NoToolDefined("No tool found in the python script.")
+    elif num_tools > 1:
+        tool_names = ", ".join(t.__name__ for t in functions + methods)
+        raise MultipleToolsDefined(f"Expected 1 but collected {num_tools} tools: {tool_names}.")
+    if functions:
+        return functions[0], None
+    else:
+        return methods[0][0], methods[0][1]
```
thy09, line 209: It is hard to understand methods[0][0] and methods[0][1]

src/promptflow/promptflow/_core/tools_manager.py
```diff
@@ -163,6 +163,33 @@ def _load_package_tool(tool_name, module_name, class_name, method_name, node_inp
...
         return getattr(provider_class(**init_inputs_values), method_name), init_inputs
 
+    @staticmethod
+    def _load_script_tool(module, tool_name, class_name, method_name, node_inputs: Mapping[str, InputAssignment]):
+        """Load script in tool with given import path and node inputs."""
+        if class_name is None:
+            return getattr(module, method_name), {}
+        provider_class = getattr(module, class_name)
+        # Note: v -- type is InputAssignment
+        init_inputs = provider_class.get_initialize_inputs()
```
thy09, line 173: Seems similar to package tool loading? Can we share some code so we could have more clear UT?

src/promptflow/promptflow/_core/tools_manager.py
```diff
@@ -308,8 +335,8 @@ def load_tool_for_script_node(self, node: Node) -> Tuple[Callable, Tool]:
         m = load_python_module_from_file(self._working_dir / path)
         if m is None:
             raise CustomToolSourceLoadError(f"Cannot load module from {path}.")
-        f = collect_tool_function_in_module(m)
-        return f, _parse_tool_from_function(f)
+        f, initialize_inputs = collect_tool_function_with_init_inputs_in_module(m)
+        return m, _parse_tool_from_function(f, initialize_inputs)
```
thy09, line 339: Why do we return module here?

src/promptflow/tests/test_configs/flows/script_tool_with_init/script_tool_with_init.py
```diff
@@ -0,0 +1,18 @@
...
+from promptflow.contracts.types import PromptTemplate
+
+
+class ScriptToolWithInit(ToolProvider):
+    def __init__(self, connection: AzureOpenAIConnection):
+        super().__init__()
+        self.connection = connection
+
+    @tool
+    def call(self, api: str, prompt: PromptTemplate, **kwargs):
```
thy09, line 14: Let's write a simpler case without connection and prompt template

src/promptflow/promptflow/_core/tool_meta_generator.py
```diff
@@ -184,6 +194,21 @@ def collect_tool_function_in_module(m):
     return tools[0]
 
 
+def collect_tool_function_with_init_inputs_in_module(m):
+    functions = collect_tool_functions_in_module(m)
```
thy09, line 198: `function_tools` and `method_tools`?

src/promptflow/promptflow/_core/tool_meta_generator.py
lumoslnt: Refined.

src/promptflow/promptflow/_core/tools_manager.py
lumoslnt: Because we need to load function from module. I have removed the module here and use a real module name when load it rather than a hard code name "__pf__main__", in order to load module based on module name latter.

src/promptflow/tests/test_configs/flows/script_tool_with_init/script_tool_with_init.py
lumoslnt: Refined.

src/promptflow/promptflow/_core/tool_meta_generator.py
lumoslnt: Refined.

src/promptflow/promptflow/_core/tool_meta_generator.py
```diff
@@ -221,6 +232,21 @@ def collect_tool_function_in_module(m):
     return tools[0]
 
 
+def collect_script_tools(m):
+    functions = collect_tool_functions_in_module(m)
+    methods = collect_tool_methods_with_init_inputs_in_module(m)
+    num_tools = len(functions) + len(methods)
+    if num_tools == 0:
+        raise NoToolDefined("No tool found in the python script.")
```
codingrabbitt1, line 240: Please redefine the error message refer to collect_tool_function_in_module func.

src/promptflow/promptflow/_core/tool_meta_generator.py
```diff
@@ -221,6 +232,21 @@ def collect_tool_function_in_module(m):
     return tools[0]
 
 
+def collect_script_tools(m):
```
codingrabbitt1, line 235: Why don't we use this function to replace collect_tool_function_in_module completely?

codingrabbitt1 (approved):

src/promptflow/promptflow/_core/tools_manager.py
```diff
@@ -233,15 +233,20 @@ def load_builtin(
...
-        return BuiltinsManager._load_package_tool(tool.name, tool.module, tool.class_name, tool.function, node_inputs)
+        return BuiltinsManager._load_tool_from_module(
+            tool.name, tool.module, tool.class_name, tool.function, node_inputs
+        )
 
     @staticmethod
-    def _load_package_tool(tool_name, module_name, class_name, method_name, node_inputs: Mapping[str, InputAssignment]):
-        """Load package in tool with given import path and node inputs."""
-        m = importlib.import_module(module_name)
+    def _load_tool_from_module(
```
thy09, line 241: I guess we may have two methods
`_load_tool_from_module` which only accepts a python module object
`_load_package_tool` which only  accepts module_name and calls `_load_tool_from_module`

The function interface is not that clean to have both `module_name` and `module`

thy09 (approved):

PR #592, opened 2023-09-22
Store package and package version for custom strong type connection, add test and support connection.key

16oeahr:
# Description

This PR does the following things:
1. Enable test for custom strong type connection, including connection CRUD, basic flow test&run.
2. Store package and package version of the connection module in order for the local vs code extension to show package info for each created connection.
3. Support user script to use their custom connection configs like **connection.key**
4. Provide refresh_connections_dir in order for the vs code extension to generate and refresh the connections directory, so that ui can show the supported types in current env

### DIFF
--- a/src/promptflow/promptflow/_sdk/entities/_connection.py
+++ b/src/promptflow/promptflow/_sdk/entities/_connection.py
@@ -673,6 +673,13 @@ class CustomStrongTypeConnection(_Connection):
         self.package_version = kwargs.get(CustomStrongTypeConnectionConfigs.PACKAGE_VERSION, None)
 
     def __getattribute__(self, item):
+        # Note: The reason to overwrite __getattribute__ instead of __getattr__ is as follows:
+        # Custom strong type connection is written this way:
+        # class MyCustomConnection(CustomStrongTypeConnection):
+        #     api_key: Secret
+        #     api_base: str = "This is a default value"
+        # api_base has a default value, my_custom_connection_instance.api_base would not trigger __getattr__.
+        # The default value will be returned directly instead of the real value in configs.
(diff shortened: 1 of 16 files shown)


chjinche (dismissed):

src/promptflow/promptflow/_sdk/entities/_connection.py
```diff
@@ -669,6 +669,26 @@ def __init__(
         super().__init__(configs=configs, secrets=secrets, **kwargs)
         self.module = kwargs.get("module", self.__class__.__module__)
         self.custom_type = custom_type or self.__class__.__name__
+        self.package = kwargs.get(CustomStrongTypeConnectionConfigs.PACKAGE, None)
+        self.package_version = kwargs.get(CustomStrongTypeConnectionConfigs.PACKAGE_VERSION, None)
+
+    def __getattribute__(self, item):
```
brynn-code, line 675: Shall we override the __getattribute__, maybe __getattr__? getattribute is more dangerous than getattr, as getattr will be called when attribute is missing, it's a fall back op, but getattribute will be called every time when access an attribute.

16oeahr: We have to call getattribute in this case.. Since we write the function this way:
```
class FakeClass:
field_1: str
field_2: str 
```

FakeClass.field_1 will not enter **getattr**

16oeahr: added comment over the function

brynn-code (approved):

chjinche (approved):

brynn-code (approved):

PR #649, opened 2023-09-27
Add error handling for is_custom_strong_type function

chenslucky:
# Description
Error message: 
image

Test result:
image

# All Promptflow Contribution checklist:
- [x] **The pull request does not introduce [breaking changes].**
- [ ] **CHANGELOG is updated for new features, bug fixes or other significant changes.**
- [x] **I have read the [contribution guidelines](../CONTRIBUTING.md).**
- [ ] **Create an issue and link to the pull request to get dedicated review from promptflow team. Learn more: [suggested workflow](../CONTRIBUTING.md#suggested-workflow).**

## General Guidelines and Best Practices
- [x] Title of the pull request is clear and informative.
- [ ] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR. For more information on cleaning up the commits in your PR, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).

### Testing Guidelines
- [ ] Pull request includes test coverage for the included changes.

PR #676, opened 2023-10-07
Refine type check for is_custom_strong_type

16oeahr:
# Description

Refine the type check for is_custom_strong_type static method as a follow-up to this pull request: https://github.com/microsoft/promptflow/pull/649

For generic type annotations like list[str], tuple[str, int],  dict[str, typing.Any] would be types.GenericAlias so return False.
For other types that would be passed to issubclass to check if it is the subclass of CustomStrongTypeConnection.

The try-except block is still kept since it cannot be confidently removed due to the uncertainty of TypeError that may occur.
TypeError is not expected to happen, but if it does, we will log it for debugging and return False.

Apart from ut, e2e flow test using below script can pass:
```
  def test(self):
        from promptflow._sdk._pf_client import PFClient

        client = PFClient()
        client.flows.test(flow=r"D:\proj\github\ms\promptflow\examples\flows\standard\gen-docstring")
```
image


src/promptflow/promptflow/contracts/tool.py
```diff
@@ -185,14 +188,28 @@ def is_connection_value(val: Any) -> bool:
...
+        # Check for instances of GenericAlias (for parameterized generic types like list[str])
+        if isinstance(val, types.GenericAlias):
+            return False
+
         try:
             return issubclass(val, CustomStrongTypeConnection)
-        except Exception:
+        except TypeError as e:
+            # TypeError is not expected to happen, but if it does, we will log it for debugging and return False.
+            # The try-except block cannot be confidently removed due to the uncertainty of TypeError that may occur.
```
chjinche, line 211: Why not capture other type exception and log?

16oeahr: According to this doc, it seems the only exception could be TypeError:
image
https://docs.python.org/3/library/functions.html

And in our promptflow code, nowhere overrides " __subclasscheck__()", so I think there coud be no other exception type

chjinche (approved):

DaweiCai (approved):

PR #642, opened 2023-09-27
Make error message format more robust.

huaiyan:
# Description

If we set parameter does not align with `message_format`, show <Param> to customer. For now, we will throw KeyError, which is one bad experience for customer.

For example:
```
ex = PromptflowException(
        message_format="Test exception message with parameters: {param}, {param1}.",
        param="test_param",
)
```
Will raise KeyError for ex.message, since we don't have `param1`.

After this PR, ex.message = "Test exception message with parameters: test_param, <param>."
# All Promptflow Contribution checklist:
- [X] **The pull request does not introduce [breaking changes].**
- [ ] **CHANGELOG is updated for new features, bug fixes or other significant changes.**
- [X] **I have read the [contribution guidelines](../CONTRIBUTING.md).**
- [ ] **Create an issue and link to the pull request to get dedicated review from promptflow team. Learn more: [suggested workflow](../CONTRIBUTING.md#suggested-workflow).**

## General Guidelines and Best Practices
- [X] Title of the pull request is clear and informative.
- [X] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR. For more information on cleaning up the commits in your PR, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).

### Testing Guidelines
- [X] Pull request includes test coverage for the included changes.

PR #680, opened 2023-10-07
Filter out None for exceptions's message_parameter

huaiyan:
`string.Formatter().parse(message_format)` will generate None in field_name.
Filter it out since it may introduce issue for jsonify.

Another option is using regular expressions,
Here's a comparison of both options from gpt:
Using string.Formatter().parse():
Pros:
- The built-in string.Formatter().parse() method is designed specifically for parsing format strings, so it is likely to handle edge cases correctly.
- This method doesn't require an additional import.

Cons:
- The code might be less readable for people who are not familiar with the string.Formatter() class and its .parse() method.
- It requires an additional function (iter_field_name()) and a loop to filter out the None values.

Using re (regular expressions):
Pros:
- The code is more concise and easier to read for those familiar with regular expressions.
- It requires fewer lines of code and directly extracts the field names, so no additional function or loop is needed for filtering None values.

Cons:
- Regular expressions might not catch all edge cases specific to format strings (although it will work correctly for most common cases).
- People not familiar with regular expressions might find this approach less readable.

In summary, both methods will work for most common cases. However, if you are mainly concerned with readability and conciseness, the regular expression method is a better choice. On the other hand, if you prioritize **handling all edge cases related to format strings**, the string.Formatter().parse() method is the safer option.

Related PR: https://github.com/microsoft/promptflow/pull/642
- [X] **The pull request does not introduce [breaking changes].**
- [X] **I have read the [contribution guidelines](../CONTRIBUTING.md).**

- [X] Title of the pull request is clear and informative.
- [X] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR. For more information on cleaning up the commits in your PR, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).

### Testing Guidelines

### DIFF
--- a/src/promptflow/tests/executor/unittests/executor/test_exceptions.py
+++ b/src/promptflow/tests/executor/unittests/executor/test_exceptions.py
@@ -12,3 +12,4 @@ class TestExceptions:
         )
 
         assert ex.message == "Test exception message with parameters: test_param, <param1>."
+        assert None not in ex.message_parameters
--- a/src/promptflow/promptflow/exceptions.py
+++ b/src/promptflow/promptflow/exceptions.py
@@ -180,7 +180,8 @@ class PromptflowException(Exception):
                 return
 
             for _, field_name, _, _ in string.Formatter().parse(message_format):
-                yield field_name
+                if field_name is not None:
+                    yield field_name
 
         # Use set to remove duplicates
         return set(iter_field_name())
--- a/src/promptflow/promptflow/exceptions.py
+++ b/src/promptflow/promptflow/exceptions.py
@@ -180,6 +180,7 @@ class PromptflowException(Exception):
                 return
 
             for _, field_name, _, _ in string.Formatter().parse(message_format):
+                # Last one field_name is always None, filter it out to avoid exception.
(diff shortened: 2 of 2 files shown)


src/promptflow/tests/executor/unittests/executor/test_exceptions.py
```diff
@@ -12,3 +12,4 @@ def test_exception_message(self):
         )
 
         assert ex.message == "Test exception message with parameters: test_param, <param1>."
+        assert None not in ex.message_parameters
```
zzn2, line 15: We should put this line in a new test case named `test_get_arguments_from_message_format`.

zzn2 (approved):

Bowen-Guo (approved):

src/promptflow/tests/executor/unittests/executor/test_exceptions.py
huaiyan: Good suggestion. Since the code is very simple now, may improve it in future.

PR #751, opened 2023-10-13
[Open_Source_LLM] Update Tool for Ignite - LLaMa2 Improvements

gjwoods:
# Description

This Pull Request improves on the success of the Open_Source_LLM with quality of life improvements and a simplified interface. More details:

1. Support providing an endpoint_name in addition to the current offering of a selecting a connection from dropdown. This text box is a temporary option, until the dynamic list functionality is available.
2. Move the priority LLaMa2 model parameters to the 1st class citizens on the Tool UX
3. Move key advanced LLaMa2 model parameters to on the Tool UX, under the Advanced Parameters section.
4. Update tests to match this experience.
5. Clean up previous code.

Here is the current experience in the VsCode experience:
image

### Testing Guidelines


src/promptflow-tools/promptflow/tools/open_source_llm.py
```diff
@@ -76,6 +73,55 @@ def format_generic_response_payload(output: bytes, response_key: str) -> str:
...
+    elif model.startswith(ModelFamily.DOLLY.lower()):
+        return ModelFamily.DOLLY
+    elif model.startswith("gpt2"):
+        return ModelFamily.GPT2
+    else:
+        raise ValueError(f"Unexpected model type: {model} derived from deployed model: {deployment_model}")
+
+
+def get_deployment_from_endpoint(endpoint_name: str, deployment_name: str = None) -> Tuple[str, str, str]:
+    from promptflow._cli._utils import get_client_for_cli
```
gjwoods, line 94: @chjinche is this the correct way to get the Workspace triple?

src/promptflow-tools/promptflow/tools/yamls/open_source_llm.yaml
```diff
@@ -6,20 +6,39 @@ promptflow.tools.open_source_llm.OpenSourceLLM.call:
   class_name: OpenSourceLLM
   function: call
   inputs:
+    endpoint_name:
+      default: "-- please enter an endpoint name --"
```
gjwoods, line 10: @chjinche is there a way I can set the length of this text box?

chjinche: add @naiyunzhang @daixianfeng to take a look.

chjinche: Offline synced, unfortunately, seems no contract way to support text box size adjustment. We will take this ask into consideration in custom tool cap design.

src/promptflow-tools/promptflow/tools/open_source_llm.py
chjinche: @brynn-code @eniac871 for CLI ask.

brynn-code: @gjwoods That's the function used by promptflow CLI, due to credentials and flexibilities concern, we suggest you creating MLClient directly by:
```
MLClient(
        credential=cred,
        subscription_id=subscription_id,
        resource_group_name=resource_group_name,
        workspace_name=workspace_name,
    )
```
Then you are able to control the workspace triple source, from tool input or environment variable, or anywhere else you like.
Same as the credential, you are able to control it by yourselves, we suggest using 'DefaultAzureCredential' here as it's a chained credential, it will try a series of credentials. (Note that interactive browser credential is excluded by default, if it is required to pop the auth window for user to login when no other available credentials, use it like:
```
MLClient(
        credential=DefaultAzureCredential(exclude_interactive_browser_credential=False),
        subscription_id=subscription_id,
        resource_group_name=resource_group_name,
        workspace_name=workspace_name,
    )
```

src/promptflow-tools/promptflow/tools/open_source_llm.py
```diff
@@ -354,58 +386,75 @@ def __call__(
...
+        model_kwargs["temperature"] = temperature
+        model_kwargs["max_new_tokens"] = max_new_tokens
+
         content_formatter = ContentFormatterFactory.get_content_formatter(
             model_family=self.model_family,
             api=api,
             chat_history=prompt
         )
 
         llm = AzureMLOnlineEndpoint(
```
prakharg-msft, line 453: @chjinche , How does telemetry work here? We are interested in capturing few metrics like
- number of times this tool got called - success and failure rate
- number  of token used from this tool, model family choosen and chat/completion
- number  of prompt flow this tool is used in and sub/ws details
- number  of deployed prompt flow that are using this tool and sub/ws details.

src/promptflow-tools/promptflow/tools/open_source_llm.py
prakharg-msft: @brynn-code, does runtime always set the ws triple in the env variables? Can we always read it from there? Is there any scenario where we would want this info to be passed down as tool input?

chjinche (approved):

src/promptflow-tools/promptflow/tools/open_source_llm.py
```diff
@@ -76,6 +77,95 @@ def format_generic_response_payload(output: bytes, response_key: str) -> str:
...
+
+    try:
+        from azure.ai.ml import MLClient
+        ml_client = MLClient(
+            credential=credential,
+            subscription_id=os.getenv("AZUREML_ARM_SUBSCRIPTION"),
+            resource_group_name=os.getenv("AZUREML_ARM_RESOURCEGROUP"),
+            workspace_name=os.getenv("AZUREML_ARM_WORKSPACE_NAME"))
+    except:
+        from promptflow.runtime import PromptFlowRuntime
```
thy09, line 109: It is not a good idea to expose "PromptFlowRuntime" here since it is an internal concept, I think we may also try credential + environment variables.

src/promptflow-tools/promptflow/tools/open_source_llm.py
```diff
@@ -76,6 +77,95 @@ def format_generic_response_payload(output: bytes, response_key: str) -> str:
...
+        return ModelFamily.GPT2
+    else:
+        raise ValueError(f"Unexpected model type: {model} derived from deployed model: {deployment_model}")
+
+
+def get_deployment_from_endpoint(endpoint_name: str, deployment_name: str = None) -> Tuple[str, str, str]:
+    from azure.identity import DefaultAzureCredential
+    credential = DefaultAzureCredential(exclude_interactive_browser_credential=False)
+
+    try:
```
thy09, line 101: We should also consider the scenario that the user doesn't have any azure bindings locally, then we may fallback to connection?

src/promptflow-tools/promptflow/tools/open_source_llm.py
```diff
@@ -76,6 +77,95 @@ def format_generic_response_payload(output: bytes, response_key: str) -> str:
...
+        from azure.ai.ml import MLClient
+        ml_client = MLClient(
+            credential=credential,
+            subscription_id=os.getenv("AZUREML_ARM_SUBSCRIPTION"),
```
chjinche, line 105: Who will set the os env?

chjinche (changes requested):

src/promptflow-tools/promptflow/tools/open_source_llm.py
chjinche: this is also why tool tests - open source llm unit tests failed. pf-tools does not depend on runtime.

src/promptflow-tools/promptflow/tools/open_source_llm.py
chjinche: for portal support, please kindly also read ws triple from env.

src/promptflow-tools/promptflow/tools/open_source_llm.py
chjinche: Could lines 109-110 be removed? It seems both local and portal versions could work via env vars. If env var approach fails, import runtime also wouldn't work. Might clean up code to just rely on env var config.

src/promptflow-tools/promptflow/tools/open_source_llm.py
gjwoods: We're trying to avoid many users from having to create a connection.
The idea here is to allow Customers not to have to setup a connection - rather than fall back we can return an error.

src/promptflow-tools/promptflow/tools/open_source_llm.py
gjwoods: They would have to - I'm not aware of any other way to get the workspace details, do you know of any @chjinche?

src/promptflow-tools/requirements.txt
```diff
@@ -1,2 +1,2 @@
 google-search-results==2.4.1
-promptflow
+promptflow[azure]
```
chjinche, line 2: This modification is inappropriate, pf-tools has no dependency on azure.

chjinche (approved):

PR #825, opened 2023-10-19
[Open_Source_LLM] Add Icon + MIR Telemetry Attribution + Fix Tests

gjwoods:
# Description
Add icon of open source llm, and attribute MIR telemetry, and fix unit tests.

### Testing Guidelines


src/promptflow-tools/tests/conftest.py
```diff
@@ -45,6 +45,24 @@ def serp_connection():
     return ConnectionManager().get("serp_connection")
 
 
+def verify_custom_connection(connection: CustomConnection) -> bool:
```
chjinche, line 48: How about renaming as `verify_oss_llm_custom_connection` to be more specific? An arbitrary custom connection may not have `endpoint_url`, then may go to line 63 `Task Succeeded unexpectedly.`

src/promptflow-tools/tests/conftest.py
```diff
@@ -45,6 +45,24 @@ def serp_connection():
...
+    from urllib.request import HTTPError
+    from urllib.error import URLError
+
+    try:
+        urllib.request.urlopen(
+            urllib.request.Request(connection.configs['endpoint_url']),
+            timeout=50)
+    except HTTPError as e:
+        # verify that the connection is not authorized, anything else would mean the endpoint is failed
+        return e.code == 403
```
chjinche, line 59: Looks if `403`, the function will return `True`, then not skip the test. Can tests run successfully if inner api call met 403?

src/promptflow-tools/tests/conftest.py
```diff
@@ -45,6 +45,24 @@ def serp_connection():
...
 
+def verify_custom_connection(connection: CustomConnection) -> bool:
+    import urllib.request
+    from urllib.request import HTTPError
+    from urllib.error import URLError
+
+    try:
+        urllib.request.urlopen(
+            urllib.request.Request(connection.configs['endpoint_url']),
+            timeout=50)
```
chjinche, line 56: add `return True` if successful call?

src/promptflow-tools/tests/conftest.py
gjwoods: Yes - we do not pass the endpoint-key here to avoid delay in generation the LLM response.

The issue I want to resolve is that it's hard to get and keep a LLaMa-2 and Falcon endpoint available for test
- while the endpoint is available (for example when I or someone else is explicitly testing code) these tests will do live E2E tests
- if the endpoint has been deleted the test will skip.

My rational for this is that if an MIR online endpoint is up (testing the URL is valid) the call to MIR will succeed when the token is passed in the tool

src/promptflow-tools/tests/conftest.py
gjwoods: No - this would mean that there is an endpoint which isn't checking authorization requirements. I image this would be a Sev 1 ICM.

chjinche: @gjwoods I see, seems a bit hard to understand without the comments, maybe you could some comments to line 53.

src/promptflow-tools/tests/conftest.py
chjinche: better to add some inline comments.

chjinche (approved):

chenslucky (approved):

PR #948, opened 2023-10-30
[Bugfix][Executor]  Execute Ndoe regardless of whether the dependencies are bypassed or not

PeiwenGaoMS:
# Description

Bug: Downstream node bypassed when dependency bypassed even when activation condition is met
Fix: When activate condition is met, the ndoe will be executed regardless of whether the dependencies are bypassed or not.

### Testing Guidelines


docs/how-to-guides/add-conditional-control-to-a-flow.md
```diff
@@ -59,7 +59,7 @@ activate:
   ![bypassed_nodes](../media/how-to-guides/conditional-flow-with-activate/bypassed_nodes.png)
 
 
-    (1) If a node has activate config and the value of `activate.when` is equals to `activate.is`, it will be bypassed.
+    (1) If a node has activate config and the value of `activate.when` is not equals to `activate.is`, it will be bypassed.
```
thy09, line 62: We should also say that it will always be executed when the condition is met.

thy09: I think the logic here is that, if activate config is set, we will always use this config.

src/promptflow/tests/test_configs/flows/activate_condition_always_met/nodeB.py
```diff
@@ -0,0 +1,8 @@
+from promptflow import tool
+
+# The inputs section will change based on the arguments of the tool function, after you save the code
```
thy09, line 3: We may rename the script to a more meaningful name.

docs/how-to-guides/add-conditional-control-to-a-flow.md
```diff
@@ -59,7 +59,7 @@ activate:
   ![bypassed_nodes](../media/how-to-guides/conditional-flow-with-activate/bypassed_nodes.png)
 
 
-    (1) If a node has activate config and the value of `activate.when` is equals to `activate.is`, it will be bypassed.
```
thy09, line 62: We may also add a note here, if we want to enforce one node always run, we can set the activate config as `dummy is dummy`
image

src/promptflow/promptflow/executor/_dag_manager.py
```diff
@@ -110,11 +110,10 @@ def _is_node_bypassable(self, node: Node) -> bool:
...
 
         # Bypass node if the activate condition is not met
-        if node.activate and (
-            self._is_node_dependency_bypassed(node.activate.condition)
-            or not self._is_condition_met(node.activate.condition, node.activate.condition_value)
-        ):
-            return True
+        if node.activate:
+            if self._is_node_dependency_bypassed(node.activate.condition):
+                return True
```
guming-llc, line 115: Not sure I understand this correctly, does this mean the node would still be bypassed if its dependency is bypassed?

thy09: Here the logic "is_node_dependency_bypassed" is checking whether the activate condition is referencing another node but the node is bypassed. It is still about activate condition, not the input dependencies.

@PeiwenGaoMS Could add more clear comments here.

PeiwenGaoMS: Added.

guming-llc: I see. It's bad naming, I'm not sure any newcomer would understand the logic in your way.

Would it be better to rename to something like "_is_referenced_node_in_condition_bypassed"?

docs/how-to-guides/add-conditional-control-to-a-flow.md
PeiwenGaoMS: Updated.

src/promptflow/tests/test_configs/flows/activate_condition_always_met/nodeB.py
PeiwenGaoMS: Updated.

src/promptflow/promptflow/executor/_dag_manager.py
PeiwenGaoMS: `_is_referenced_node_in_condition_bypassed` is inappropriate because this function is not only used to determine whether the condition has been skipped. For a node, its dependencies are not only inputs, but also the node referred in skip config and activate config.

PeiwenGaoMS: Will app some comments to this function.

guming-llc: Oh I see what you mean here, so the "dependency" in the function name is not for the "node" itself, but for the input parameter "node.activate.condition"
That's what confused me.
Maybe something like "_is_reference_node_bypassed" would be more friendly to newcomer.
Not blocker though.

guming-llc: To elaborate a little more of my confusion:
As you said "For a node, its dependencies are not only inputs, but also the node referred in skip config and activate config", and if I understand the function as "whether _is_node_dependency_bypassed for a node", then I'm confused, since we clearly deal with input referenced and condition referenced node differently.

thy09 (approved):

Hhhilulu (approved):

PR #994, opened 2023-11-06
add token provider for aoai connection

melionel:
# Description

### Testing Guidelines


src/promptflow/promptflow/_sdk/entities/_connection.py
```diff
@@ -346,10 +347,12 @@ class AzureOpenAIConnection(_StrongTypeConnection):
     TYPE = ConnectionType.AZURE_OPEN_AI
 
     def __init__(
-        self, api_key: str, api_base: str, api_type: str = "azure", api_version: str = "2023-07-01-preview", **kwargs
+        self, api_key: str, api_base: str, api_type: str = "azure", api_version: str = "2023-07-01-preview",
+        token_provider: Union[str, TokenProviderABC] = AzureTokenProvider(), **kwargs
```
thy09, line 351: Usually it would be better to have default as None and create the instance in the init function.
Otherwise a AzureTokenProvider instance would be used in different connection instances.

src/promptflow/promptflow/_sdk/entities/_connection.py
```diff
@@ -386,6 +389,14 @@ def api_version(self, value):
         """Set the connection api version."""
         self.configs["api_version"] = value
 
+    @property
+    def token(self):
+        """Return the connection token."""
+        if isinstance(self.token_provider, str):
```
thy09, line 395: It is not that easy to understand that token provider could be a string?

src/promptflow/promptflow/_sdk/entities/_connection.py
```diff
@@ -386,6 +389,14 @@ def api_version(self, value):
         """Set the connection api version."""
         self.configs["api_version"] = value
 
+    @property
+    def token(self):
```
wangchao1230, line 393: should be get_token() here also?

src/promptflow/promptflow/_sdk/entities/_connection.py
wangchao1230, line 351: User should be able to create connection with this field in yaml?

src/promptflow/promptflow/_core/token_provider.py
```diff
@@ -0,0 +1,65 @@
+import threading
+import time
+
+from abc import ABC, abstractmethod
+from azure.identity import DefaultAzureCredential
+from datetime import datetime, timedelta
+
+COGNITIVE_AUDIENCE = "https://cognitiveservices.azure.com/"
```
wangchao1230, line 8: why is this a constant? may document why is this value

src/promptflow/promptflow/_core/token_provider.py
```diff
@@ -0,0 +1,65 @@
...
+        with self._token_lock:
+            cached_token = self.audience_tokens.get(audience, None)
+            expiry_time = self.audience_expiry_times.get(audience, None)
+
+        # Check if there is a cached token and it's not expired
+        if cached_token and expiry_time and datetime.now() < expiry_time:
+            return cached_token
+
+        # If no cached token or it's expired, refresh the token
+        new_token, expiry_time = self.refresh_token(audience)
```
wangchao1230, line 48: DefaultAzureCredential should already has the token cache logic. what we need to do is to cache the credential obj.

src/promptflow/promptflow/_sdk/entities/_connection.py
melionel: Agree, but we need to provide an easy way for user to provide a static token for a connection. What's your suggestion ?

src/promptflow/promptflow/_sdk/entities/_connection.py
melionel: Yes, sounds better

src/promptflow/promptflow/_sdk/entities/_connection.py
melionel: I don't think so, and that's why the token_provider is not a field of AzureOpenAIConnectionSchema.
Do we have scenario that user needs to override the token_provider when create connection from yaml ?

src/promptflow/promptflow/_core/token_provider.py
melionel: This is the required audience for a valid token to access azure openai service.

src/promptflow/promptflow/_sdk/entities/_connection.py
thy09: I think we can have a StaticTokenProvider instead of put a string as a token provider?

melionel: Get your point, I think that's a good idea.

src/promptflow/promptflow/_sdk/entities/_connection.py
```diff
@@ -346,10 +347,12 @@ class AzureOpenAIConnection(_StrongTypeConnection):
     TYPE = ConnectionType.AZURE_OPEN_AI
 
     def __init__(
-        self, api_key: str, api_base: str, api_type: str = "azure", api_version: str = "2023-07-01-preview", **kwargs
+        self, api_key: str, api_base: str, api_type: str = "azure", api_version: str = "2023-07-01-preview",
+        token_provider: TokenProviderABC = None, **kwargs
```
wangchao1230, line 351: add this param to docstring?

thy09 (approved):

wangchao1230 (approved):

PR #1420, opened 2023-12-07
[promptflow][Test] Support async network behavior in replay test

zhengfeiwang:
# Description

Run download feature uses async network request, which is not supported in previous replay test; this PR targets to support this:

- Add a fixture for async HTTP request: a known issue for vcrpy, but the related code change has not been merged.
- Handle different contract across sync and async response.

### Testing Guidelines

### DIFF
--- a/src/promptflow/tests/sdk_cli_azure_test/recording_utilities/bases.py
+++ b/src/promptflow/tests/sdk_cli_azure_test/recording_utilities/bases.py
@@ -119,7 +119,18 @@ class PFAzureIntegrationTestRecording:
         if is_live():
             return response
 
-        response["body"]["string"] = response["body"]["string"].decode("utf-8")
+        # sync and async responses have different structure
+        # sync has .body.string, while async has .content
+        # in our sanitizers (processors) logic, we only handle .body.string
+        # so make async align sync for less code change
+        is_async_response = False
+        if "body" not in response:
+            is_async_response = True
+            body_string = response.pop("content")
+            response["body"] = {"string": body_string}
+        else:
+            response["body"]["string"] = response["body"]["string"].decode("utf-8")
+
         if is_record():
             # lower and filter some headers
             headers = {}
@@ -131,7 +142,22 @@ class PFAzureIntegrationTestRecording:
             for processor in self.recording_processors:
                 response = processor.process_response(response)
 
-        response["body"]["string"] = response["body"]["string"].encode("utf-8")
+        if is_async_response:
+            response["content"] = response["body"]["string"]
+            if not is_replay():
+                response.pop("body")
+                if isinstance(response["content"], bytes):
+                    response["content"] = response["content"].decode("utf-8")
+            else:
+                # vcrpy originally support async response
+                # but due to our fixture/patch for httpx, we need some manual work to add some fields
+                # otherwise, replay tests will break during init VCR response instance
+                response["status"] = {"code": response["status_code"], "message": ""}
+                if isinstance(response["body"]["string"], str):
+                    response["body"]["string"] = response["body"]["string"].encode("utf-8")
+        else:
+            response["body"]["string"] = response["body"]["string"].encode("utf-8")
+
         return response
 
     def _get_recording_processors(self) -> List[RecordingProcessor]:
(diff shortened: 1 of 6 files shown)


src/promptflow/tests/sdk_cli_azure_test/recording_utilities/bases.py
```diff
@@ -119,7 +119,18 @@ def _process_response_recording(self, response: Dict) -> Dict:
         if is_live():
             return response
 
-        response["body"]["string"] = response["body"]["string"].decode("utf-8")
+        # sync and async responses have different structure
+        # sync has .body.string, while async has .content
+        # in our sanitizers (processors) logic, we only handle .body.string
+        # so make async align sync for less code change
+        is_async_response = False
+        if "body" not in response:
```
0mza987, line 127: do we have a better choice to judge if this is an async response? This seems fragil

src/promptflow/tests/sdk_cli_azure_test/recording_utilities/bases.py
```diff
@@ -230,6 +256,17 @@ def _drop_duplicate_recordings(self) -> None:
         return
 
     def _custom_request_path_matcher(self, r1: Request, r2: Request) -> bool:
+        # NOTE: orders of below conditions matter, please modify with caution
+        # in run download scenario, observed below wired path: https://<xxx>/https://<yyy>/<remaining>
```
0mza987, line 260: Is this the same issue with above github issue?

zhengfeiwang: no, this is the issue we synced before, I wonder it's a blob client behavior: get the path somewhere (maybe PFS), then use it to download

src/promptflow/tests/sdk_cli_azure_test/recording_utilities/bases.py
zhengfeiwang: I think it will be better if judge from key `content`, which seems a contract for httpx, which is safer to use.

src/promptflow/tests/sdk_cli_azure_test/recording_utilities/bases.py
```diff
@@ -131,7 +142,22 @@ def _process_response_recording(self, response: Dict) -> Dict:
             for processor in self.recording_processors:
                 response = processor.process_response(response)
 
-        response["body"]["string"] = response["body"]["string"].encode("utf-8")
+        if is_async_response:
+            response["content"] = response["body"]["string"]
+            if not is_replay():
```
D-W-, line 147: Why do we need to do this?

D-W- (approved):

src/promptflow/tests/sdk_cli_azure_test/recording_utilities/bases.py
zhengfeiwang: in short, httpx response has different contract with others network pkgs, so we need some different logic to handle.

src/promptflow/tests/sdk_cli_azure_test/recording_utilities/utils.py
```diff
@@ -226,3 +234,10 @@ def is_json_payload_response(response: Dict) -> bool:
     headers = response.get("headers")
     # PFAzureIntegrationTestRecording will lower keys in response headers
     return _is_json_payload(headers, key="content-type")
+
+
+def is_httpx_response(response: Dict) -> bool:
+    # different from other stubs in vcrpy, httpx response uses "content" instead of "body"
```
0mza987, line 240: seems aiohttp also uses `content`, but it may not matter here.

0mza987 (approved):

src/promptflow/tests/sdk_cli_azure_test/recording_utilities/utils.py
zhengfeiwang: not sure, but seems `vcrpy` works fine with `aiohttp`, so I guess it can successfully handle `aiohttp` case, so I left the comment for "other stubs in vcrpy"...

D-W- (approved):

src/promptflow/tests/sdk_cli_azure_test/recording_utilities/bases.py
zhengfeiwang: in this place, non-replay mode means record mode (as live will directly return at the very beginning)
then in record mode, response need to be returned as what it looks like as httpx response, which means, string type content, and no body.string; otherwise, it will be recorded in recording file as duplication; so in this case, we need to pop body part, and promise the string type.

PR #1443, opened 2023-12-11
[SDK/CLI] Support run create from local run folder

0mza987:
# Description

- Enable `pf run create` with local exisiting run folder.
    -`pf run create --source <path-to-run-folder>`
- Refine `pfazure run download`:
   - Download metadata of run.
   - Use thread pool to run sync functions.

### Testing Guidelines


src/promptflow/promptflow/_sdk/operations/_run_operations.py
```diff
@@ -101,16 +105,35 @@ def create_or_update(self, run: Run, **kwargs) -> Run:
...
-        print("======= Run Summary =======\n")
-        duration = str(run._end_time - run._created_on)
-        print(
-            f'Run name: "{run.name}"\n'
-            f'Run status: "{run.status}"\n'
-            f'Start time: "{run._created_on}"\n'
-            f'Duration: "{duration}"\n'
-            f'Output path: "{run._output_path}"\n'
-        )
+        # print run summary if they run is not from existing run folder
```
zhengfeiwang, line 126: why we skip print summary for existing run? is there any missing information?

src/promptflow/promptflow/_sdk/_orm/run_info.py
```diff
@@ -158,3 +159,12 @@ def list(max_results: Optional[int], list_view_type: ListViewType) -> List["RunI
                 return [run_info for run_info in basic_statement.limit(max_results)]
             else:
                 return [run_info for run_info in basic_statement.all()]
+
+    @staticmethod
+    @sqlite_retry
+    def delete(name: str) -> None:
+        with mgmt_db_session() as session:
+            run_info = session.query(RunInfo).filter(RunInfo.name == name).first()
```
zhengfeiwang, line 167: as this is in ORM level operation, I prefer raise an error if `run_info` is `None`, which means the delete operation is failed/invalid.

src/promptflow/promptflow/_sdk/entities/_run.py
```diff
@@ -612,3 +635,22 @@ def _generate_output_path(self, config: Optional[Configuration]) -> Path:
...
+        source = Path(source)
+        params_override = params_override or {}
+        if not source.exists() or not source.is_dir():
+            raise UserErrorException(f"Invalid run source: {source!r}. Expecting a valid run source folder.")
+
+        return cls(
+            name=source.name,
+            source=source,
+            run_source=RunInfoSources.EXISTING_RUN,
+            status=RunStatus.COMPLETED,  # failed run shall not be created in this way
```
zhengfeiwang, line 651: if we don't expect failed run here, shall we add some status check at the very beginning of this function?

zhengfeiwang: ah, maybe it's kind of hard to do at the top as you may not know it at all, but somehow I think we might need a check?

src/promptflow/promptflow/_sdk/operations/_run_operations.py
0mza987: Good catch, currently the downloaded run info has no run meta info like start time and endtime, so the time calculation here is not invalid. 

The flow artifacts have child run's start time and end time, I'm not sure if it's OK to pick the start time of first run and end time of the last run and regard it as the duration of the run record.

src/promptflow/promptflow/_sdk/entities/_run.py
0mza987: It's considered, but the problem here is we don't know the status of this run, this run folder can be downloaded or simply shared. For the download case, maybe we can write some additional info in the run metadata like start/end time, for the later case,there is no good solution yet.

src/promptflow/promptflow/_sdk/entities/_run.py
```diff
@@ -196,18 +202,32 @@ def properties(self) -> Dict[str, str]:
...
+        # 1. LOCAL: Created when run is created from local flow
+        # 2. EXISTING_RUN: Created when run is created from existing run folder
+        source = None
+        if getattr(obj, "run_source", None) == RunInfoSources.EXISTING_RUN:
+            source = properties_json[FlowRunProperties.OUTPUT_PATH]
+
         return Run(
             name=str(obj.name),
-            flow=Path(properties_json[FlowRunProperties.FLOW_PATH]),
+            flow=Path(flow) if flow else None,
```
D-W-, line 229: Can we directly use snapshot path as flow path when flow's from source?

0mza987: In content wise they are equal, but in concept wise the flow path is the flow that the run was created against and snapshot is a copy of that flow. Pointing the flow path to a snapshot folder is kinda twisted, so I'm not sure if we should do that.

src/promptflow/promptflow/_sdk/schemas/_run.py
```diff
@@ -60,7 +60,10 @@ class RunSchema(YamlFileSchema):
     properties = fields.Dict(keys=fields.Str(), values=fields.Str(allow_none=True))
     # endregion: common fields
 
-    flow = UnionField([LocalPathField(required=True), fields.Str(required=True)])
+    # one of ["flow", "source"] must be specified, it will be validated through function `validate_run_source`
+    flow = UnionField([LocalPathField(), fields.Str()])
+    source = UnionField([LocalPathField(), fields.Str()])
```
D-W-, line 65: Not sure if we want to expose source in run.yml

src/promptflow/promptflow/_sdk/operations/_run_operations.py
0mza987: Recovered, now we have start time and end time.

wangchao1230 (approved):

zhengfeiwang (approved):

PR #1582, opened 2023-12-25
[PromptFlow][Executor][Internal] Refine process pool. 

Hhhilulu:
# Description
This pull request is mainly to solve the deadlock problem caused by the current process pool.

# Previous Design
**Create process in each thread in the thread pool. (This causes deadlock in fork mode, it’s best not to create process in thread.)**

# Solution
Remove the process creation from the thread pool and create the child process in a clean environment.
For fork mode
1. Spawn a subprocess in main process to establish a clean environment.
2. Fork [worker_count] sub-process in this environment to execute batch run.
3. Create thread pool of size [worker_count] and monitor the above sub-process.
For spawn mode
1. When using spawn, create [worker_count] subprocess in main process.

# Main changes

1. [src/promptflow/promptflow/executor/_line_execution_process_pool.py](https://github.com/microsoft/promptflow/blob/48e565bb4d493b5132c8a5dbb8cfa5d83cad7861/src/promptflow/promptflow/executor/_line_execution_process_pool.py)

Create [worker_count] input_queues, output_queues, process_info_dict and control_signal_queue in main_process and pass them to sub-process.

- Input_queues

Each fork sub-process corresponds to an input_queue, used to receive the input of batch run from main process. 

- Output_queues:

Each fork sub-process corresponds to an output_queue, use to receive the line result of the batch run and pass result to the main process.

- Control_signal_queue:

Used to send signals from main process to fork sub-procoss, such as, ‘restart’, ‘end’, ’start’.

- Process_info_dict:

Use to save [worker_count] fork sub-processes’ information, the process info’s key is index, the value is process id and process name which will use to print some logs and monitor process’s status, alive or crash.

2. [src/promptflow/promptflow/executor/_process_manager.py](https://github.com/microsoft/promptflow/blob/48e565bb4d493b5132c8a5dbb8cfa5d83cad7861/src/promptflow/promptflow/executor/_process_manager.py)

- Defines a base class AbstractProcessManager and three implementation classes ForkProcessManager, SpawnProcessManager and SpawnedForkProcessManager.
- Each class implements new_process, end_process and restart_process functions. The start_processes method is added to the implementation class to start the [work_conut] sub-process. 
- The differences between the three functions in different implementation classes:
  - **SpawnProcessManager**:
    - The three functions directly operate on the process object.
  - **ForkProcessManager**:
    - Just put the ‘start’, ‘end’ and ‘restart’ signals to the control_signal_queue. The real implementation is completed in the sub fork process, that is SpawnedForkProcessManager, which is mainly manage the fork sub process and handle signals. 
    - In ForkProcessManager, the start_processes will first spawn a child process and then fork [work_count] sub-process, but in SpawnProcessManager, the start_processes will directly spawn [worker_count] sub-process.

Test case:
[portal test](https://ml.azure.com/prompts/flow/f8925d22-478a-4b5c-8e16-a85158d03b9b/ae12615d-85bb-4fd8-826a-8e14ff0011d9/run/a98c4b1f-fdb9-467f-9284-f3d1a2871735/details?wsid=/subscriptions/96aede12-2f73-41cb-b983-6d11a904839b/resourceGroups/promptflow/providers/Microsoft.MachineLearningServices/workspaces/promptflow-eastus-dev&flight=promptflow&tid=72f988bf-86f1-41af-91ab-2d7cd011db47)
   
# Aditional changes

- Add log_process unit test and e2e test.
- Add handle process crashes.
- Add environment variable PF_LINE_TIMEOUT_SEC.
- Change _DEFAULT_WORKER_COUNT from 16 to 4.
- Remove legacy code.

### Testing Guidelines


src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -120,19 +106,119 @@ def put(self, args):
...
+        )
+    else:
+        bulk_logger.info(
+            f"Process name: {process_name}, Process id: {pid}, Line number: {line_number} start execution."
+        )
+
+    return f"Process name({process_name})-Process id({pid})-Line number({line_number})"
+
+
+def create_process_fork(
```
guming-learning, line 123: Seems the functionality is not just create process with fork, it actually manage a fork process manager,
 probably call it "fork_processes_manager"

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -196,6 +283,20 @@ def __enter__(self):
 
         self._inputs_queue = Queue()
         self._n_process = self._determine_worker_count()
+        if self._use_fork:
+            self._input_queues = [manager.Queue() for i in range(self._n_process)]
+            self._output_queues = [manager.Queue() for i in range(self._n_process)]
+            self._restart_queue = manager.Queue()
```
guming-learning, line 289: Need to change it to sth like "_control_signal_queue", since it does both deletion and restarting.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -120,19 +106,130 @@ def put(self, args):
...
+        )
+    else:
+        bulk_logger.info(
+            f"Process name: {process_name}, Process id: {pid}, Line number: {line_number} start execution."
+        )
+
+    return f"Process name({process_name})-Process id({pid})-Line number({line_number})"
+
+
+def fork_processes_manager(
```
guming-learning, line 123: If it's convenient, please add some comment to explain the current process model, and the reason of the current design.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -120,19 +106,130 @@ def put(self, args):
...
+                executor_creation_func,
+                input_queues[i],
+                output_queues[i],
+                log_context_initialization_func,
+                current_operation_context
+            ),
+            daemon=True
+        )
+        process.start()
+        input_queues[i].put((i, process.pid, process.name))
```
guming-learning, line 160: We can see that it's not obvious what does "input_queues" do and why we put a tuple (i, process.pid, procees.name) to it, please add some annotation to explain the purpose of input/output queues or use a longer name to show it's real purpose

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -207,66 +324,171 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+                    # Monitor process aliveness and start new one if it crashes
+                    if self._use_fork:
+                        if not psutil.pid_exists(pid):
+                            # If the process crashes, set the 'restart_outer_loop' to TRUE.
+                            # And re-execute the task from the beginning.
+                            restart_outer_loop = True
+                            # Clear the contents of input_queue to allow its reuse in fork mode.
+                            input_queue.get()
+                            # Put unfinished tasks into task_queue again.
+                            task_queue.put(args)
```
guming-learning, line 376: I think we don't need this put back logic.
If the process crashes, we just handle it as failed.

You have already put the `args` into input_queue in line 358, if the process crashes after it get task from input queue, then `input_queue.get()` will stuck.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -120,19 +106,130 @@ def put(self, args):
...
+        control_signal_queue,
+        flow_file,
+        connections,
+        working_dir,
+        raise_ex,
+):
+    signal.signal(signal.SIGINT, signal_handler)
+    process_info = {}
+    context = multiprocessing.get_context("fork")
+    run_storage = QueueRunStorage(output_queues[0])
```
thy09, line 137: What is "output_queues[0]"?

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -120,19 +106,130 @@ def put(self, args):
...
+        process.start()
+        input_queues[i].put((i, process.pid, process.name))
+        process_info[process.pid] = {'process': process}
+        return process
+
+    for i in range(len(input_queues)):
+        new_process(i)
+
+    def kill_and_remove_process(pid):
+        process = process_info[pid]['process']
```
thy09, line 168: Not a good idea to use a dict with key `process` to get the value, maybe using a dataclass.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -120,19 +106,130 @@ def put(self, args):
...
+        )
+        process.start()
+        input_queues[i].put((i, process.pid, process.name))
+        process_info[process.pid] = {'process': process}
+        return process
+
+    for i in range(len(input_queues)):
+        new_process(i)
+
+    def kill_and_remove_process(pid):
```
thy09, line 167: There are many inner functions, maybe we can have a class to do such stuffs instead of inner classes.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -197,6 +297,23 @@ def __enter__(self):
         self._inputs_queue = Queue()
         self._n_process = self._determine_worker_count()
 
+        if self._use_fork:
```
thy09, line 300: It is tricky that `use_fork` and `not use_fork` are in different places, is it necessary?

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -207,66 +324,171 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    def _timeout_process_wrapper(
+            self,
+            run_start_time: datetime,
+            task_queue: Queue,
+            timeout_time,
+            result_list,
+            input_queue=None,
+            output_queue=None
+    ):
+        if self._use_fork:
```
thy09, line 474: It may not be a good practice to have a lot of `self._user_fork` in different places.
We may have two choices:
1. If we can have most logic common, we may try put the difference only when initializing the processes, then use the same interface to consume input data.
2. If most logic cannot be shared, we may have two different classes to handle the two scenarios.

Personally I prefer 1, but it seems that the current implementations have a lot of fork specific logics.

thy09: I think we'd be better revisit the interface design before do the implementation.

thy09: Just a simple proposal according to this PR:
```
class AbstractProcessManager:
    def create_process(idx):
        pass
    def restart_process(idx):
        pass
    
For SpawnProcessManager, here we just simply start and stop processes.
For ForkProcessManager, here we will send a signal to the spawned main process for creation, and send another signal to the spawned process for stop and restart.
```

thy09: We may add more interfaces to handle the task consumption and timeout monitoring.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: Get (i, process.pid, procees.name) through process_info dict, no longer from input_queue

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: When the process crashed, set this line to failed

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: Removed

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: Use process_info manager.dict to transfer information between processes

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: Add ProcessManager, ForkProcessManager, SpawnProcessManager

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_process_manager.py
```diff
@@ -0,0 +1,219 @@
...
+import queue
+import signal
+from functools import partial
+import psutil
+from promptflow._core.operation_context import OperationContext
+from promptflow._utils.logger_utils import LogContext, bulk_logger
+from promptflow.executor.flow_executor import FlowExecutor
+
+
+class ProcessManager:
```
thy09, line 11: It would be better to add more document string to explain the usage of the classes.

src/promptflow/promptflow/executor/_errors.py
```diff
@@ -175,6 +175,15 @@ def __init__(self, line_number, timeout):
         )
 
 
+class ProcessCrashError(UserErrorException):
+    """Exception raised when single line execution timeout"""
```
thy09, line 179: The name is crash error but the description is execution timeout?

src/promptflow/promptflow/executor/_process_manager.py
```diff
@@ -0,0 +1,327 @@
...
+    '''
+    AbstractProcessManager is a base class for managing processes.
+
+    :param input_queues: Queues for providing input data to the processes.
+    :type input_queues: List[multiprocessing.Queue]
+
+    :param output_queues: Queues for receiving execution results of the processes.
+    :type output_queues: List[multiprocessing.Queue]
+
+    :param process_info: Dictionary to store information about the processes.
```
thy09, line 23: It would be more clear to use a dataclass instead of a dictionary for process info.

src/promptflow/promptflow/executor/_process_manager.py
```diff
@@ -0,0 +1,327 @@
...
+    :type process_info: dict
+
+    :param process_target_func: The target function that the processes will execute.
+
+    :param raise_ex: Flag to determine whether to raise exceptions or not.
+    :type control_signal_queue: multiprocessing.Queue
```
thy09, line 29: type is signal queue? confused.

src/promptflow/promptflow/executor/_process_manager.py
```diff
@@ -0,0 +1,327 @@
...
+        flow_file=flow_file,
+        connections=connections,
+        working_dir=working_dir,
+        raise_ex=raise_ex,
+    )
+
+    executor_creation_func = partial(create_executor_fork, flow_executor=executor)
+
+    # Function to create and start a new process.
+    def new_process(i):
```
thy09, line 262: We might extract these logic in another new class `SpawnedForkProcessManager`, then the logic could be more clear like:
```
def fork_process_manager():
    manager = SpawnedForkProcessManager()
    for i in range(len(input_queues)):
        manager.new_process(i)
    while True:
        # handle signals
```

src/promptflow/promptflow/executor/_errors.py
Hhhilulu: Oh, sorry, copied from others and fixed

src/promptflow/promptflow/executor/_process_manager.py
Hhhilulu: Sorry, copied from others and fixed

src/promptflow/promptflow/executor/_process_manager.py
```diff
@@ -0,0 +1,327 @@
...
+            process = psutil.Process(pid)
+            process.terminate()
+            process.wait()
+            process_info.pop(i)
+        except psutil.NoSuchProcess:
+            bulk_logger.warning(f"Process {pid} had been terminated")
+
+    # Function to handle control signals for processes.
+    def handle_signals(control_signal, i):
+        if control_signal == "end":
```
thy09, line 295: We may use enums for such signals.

src/promptflow/promptflow/executor/_process_manager.py
```diff
@@ -0,0 +1,327 @@
...
+        self._raise_ex = raise_ex
+
+        current_log_context = LogContext.get_current()
+        self._log_context_initialization_func = (
+            current_log_context.get_initializer() if current_log_context else None)
+        self._current_operation_context = OperationContext.get_instance().get_context_dict()
+
+    def new_process(self, i):
+        # i: Index of the new process to start.
+        pass
```
thy09, line 53: raise NotImplementedError

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -197,6 +134,42 @@ def __enter__(self):
         self._inputs_queue = Queue()
         self._n_process = self._determine_worker_count()
 
+        self._input_queues = [manager.Queue() for _ in range(self._n_process)]
+        self._output_queues = [manager.Queue() for _ in range(self._n_process)]
+        self._control_signal_queue = manager.Queue()
+        self._process_info = manager.dict()
```
thy09, line 140: Maybe `process_info_dict` to indicate it is a dict to store process infos.
Currently it seems to be passing initialized process infos to subprocesses.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -197,6 +134,42 @@ def __enter__(self):
...
+                self._connections,
+                self._working_dir,
+                self._input_queues,
+                self._output_queues,
+                self._process_info,
+                _process_wrapper,
+                False,
+            )
+        else:
+            self._processes_manager = SpawnProcessManager(
```
thy09, line 163: Please also add comments for spawn mode.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -207,65 +180,121 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
-
+    def _process_task(
+        self,
+        process_info,
+        task_queue,
+        timeout_time,
+        result_list
+    ):
         while True:
+            index, pid, process_name, input_queue, output_queue = process_info
```
thy09, line 191: It is not a good practice to use a tuple to carry such a lot information, we may use dataclass or something to carry these.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -207,65 +180,121 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
             start_time = datetime.utcnow()
             completed = False
+            crashed = False
 
             while datetime.utcnow().timestamp() - start_time.timestamp() <= timeout_time:
                 try:
+                    # Monitor process aliveness.
+                    if not psutil.pid_exists(pid):
+                        crashed = True
+                        completed = True
```
thy09, line 212: Why completed is True?

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -207,65 +180,121 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        while True:
+            try:
+                pid = self._process_info[index]['pid']
+                process_name = self._process_info[index]['process_name']
+                process_info = (index, pid, process_name, input_queue, output_queue)
+                return process_info
+            except KeyError:
+                continue
+
+    def _process_message(self, message, result_list):
```
thy09, line 253: process_output_message?

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -207,65 +180,121 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+
+    def handle_process_crashed(self, line_number, inputs, run_id, start_time, result_list):
+        bulk_logger.warning(f"Process crashed while executing line {line_number},")
+        ex = ProcessCrashError(line_number)
+        result = self._generate_line_result_for_exception(
+            inputs, run_id, line_number, self._flow_id, start_time, ex
+        )
+        result_list.append(result)
+
+    def _timeout_process_wrapper(
```
thy09, line 281: We may have a more meaningful name of this function and better comments.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -365,20 +394,34 @@ def run(self, batch_inputs):
             ),
         ):
             try:
+
+                base_args = (self._inputs_queue, self._line_timeout_sec, result_list)
```
thy09, line 398: We'd better explain what is the meaning of such arguments.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -365,20 +394,34 @@ def run(self, batch_inputs):
...
                 try:
+                    # Only log when the number of results changes to avoid duplicate logging.
+                    last_log_count = 0
                     # Wait for batch run to complete or KeyboardInterrupt
                     while not async_result.ready():
-                        # Check every 1 second
+                        current_result_count = len(result_list)
+                        if current_result_count != last_log_count:
+                            log_progress(
+                                run_start_time=run_start_time,
```
thy09, line 418: Do we have test case to cover the logs here？

src/promptflow/promptflow/executor/_process_manager.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_process_manager.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_process_manager.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: Process_info is a dictionary shared between multiple processes, similar to input_queue and output_queue. 
It changes in real-time in the child process and gets the latest value in the main process.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_process_manager.py
```diff
@@ -0,0 +1,458 @@
...
+from enum import Enum
+import psutil
+from promptflow._core.operation_context import OperationContext
+from promptflow._utils.logger_utils import LogContext, bulk_logger
+from promptflow.executor.flow_executor import FlowExecutor
+
+
+@dataclass
+class ProcessInfo:
+    index: int = None
```
thy09, line 17: We may not have default values if the default None is not expected.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -208,65 +180,131 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
             self._pool.join()
 
-    def _timeout_process_wrapper(self, run_start_time: datetime, task_queue: Queue, timeout_time, result_list):
-        healthy_ensured_process = HealthyEnsuredProcess(self._executor_creation_func, self.context)
-        healthy_ensured_process.start_new(task_queue)
-
-        if not healthy_ensured_process.process.is_alive():
-            return
-
+    def _process_task(
```
thy09, line 183: Add docstring to explain what does "process_task" mean and how will it be called.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -208,65 +180,131 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+
                     # Responsible for checking the output queue messages and
                     # processing them within a specified timeout period.
-                    message = healthy_ensured_process.get()
-                    if isinstance(message, LineResult):
-                        completed = True
-                        message = self._process_multimedia(message)
-                        result_list.append(message)
+                    message = output_queue.get(timeout=1)
+                    completed = self._process_output_message(message, result_list)
```
thy09, line 223: It is a little confusing the function "process_output_message" returns a boolean value. Considering a more accurate function name or adjust the logic as follows:
```
self._process_output_message
if isinstance(message, LineResult)  # Add comments to explain why it is completed here
    break
```
```

File: src/promptflow/promptflow/executor/_line_execution_process_pool.py [hunk 29]
```diff
@@ -366,20 +404,44 @@ def run(self, batch_inputs):
             ),
         ):
             try:
+                args_list = [
+                    (
+                        self._task_queue,        # Shared task queue for all sub processes to read the input data.
+                        self._line_timeout_sec,  # Line execution timeout.
+                        result_list,             # Bath run result lit.
```
thy09, line 411: typo

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -208,65 +180,131 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+            task_queue: Queue,
+            timeout_time,
+            result_list,
+            index,
+            input_queue,
+            output_queue
+    ):
+        process_info = self._get_process_info(index, input_queue, output_queue)
+
+        self._process_task(
```
thy09, line 302: Seems this `_process_task` is actually `monitor_workers_and_process_tasks_in_thread`?

src/promptflow/promptflow/executor/_process_manager.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: Added

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: Remove this function and add this part to '_monitor_workers_and_process_tasks_in_thread'

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -208,65 +180,131 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+            input_queue.put(args)
             inputs, line_number, run_id = args[:3]
-            self._processing_idx[line_number] = healthy_ensured_process.format_current_process(line_number)
 
+            self._processing_idx[line_number] = format_current_process(process_name, process_id, line_number)
             start_time = datetime.utcnow()
             completed = False
+            crashed = False
 
             while datetime.utcnow().timestamp() - start_time.timestamp() <= timeout_time:
```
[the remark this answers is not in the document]
Hhhilulu: fixed

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: completed = self._handle_output_queue_messages(output_queue, result_list)
    if completed:
        break

rename the function

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: Yes, remove '_process_task' and move it in 'monitor_workers_and_process_tasks_in_thread'

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: Renamed

src/promptflow/tests/sdk_cli_test/e2etests/test_cli.py
```diff
@@ -56,6 +57,37 @@ def run_pf_command(*args, cwd=None):
...
+        f"{FLOWS_DIR}/simple_flow_with_ten_inputs/data.jsonl",
+        "--name",
+        run_id,
+    )
+    run = local_client.runs.get(name=run_id)
+    local_storage = LocalStorageOperations(run)
+    detail = local_storage.load_detail()
+    flow_runs_list = detail["flow_runs"]
+    for i, flow_run in enumerate(flow_runs_list):
+        if line_timeout_seconds == "54":
```
thy09, line 78: Using `== "54"` and ` != 9` may not be a good practice to do these check.
Maybe we can have another parameter `expected_failed_line_numbers`

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -60,84 +61,21 @@ def persist_flow_run(self, run_info: FlowRunInfo):
...
-            bulk_logger.info(
-                f"Process name: {process_name}, Process id: {process_pid}, Line number: {line_number} completed."
-            )
-        else:
-            bulk_logger.info(
-                f"Process name: {process_name}, Process id: {process_pid}, Line number: {line_number} start execution."
-            )
-
-        return f"Process name({process_name})-Process id({process_pid})-Line number({line_number})"
+    return f"Process name({process_name})-Process id({pid})-Line number({line_number})"
```
thy09, line 74: May not be a good practice to "print something in a string format function".

src/promptflow/tests/sdk_cli_test/e2etests/test_cli.py
Hhhilulu: Add time_out_index

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: Add log_process_status function to print.'
The function 'format_current_process' focus on format current process.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -60,84 +61,22 @@ def persist_flow_run(self, run_info: FlowRunInfo):
...
-                input_queue,
-                output_queue,
-                current_log_context.get_initializer() if current_log_context else None,
-                OperationContext.get_instance().get_context_dict(),
-            ),
-            # Set the process as a daemon process to automatically terminated and release system resources
-            # when the main process exits.
-            daemon=True,
-        )
+def format_current_process(process_name, pid, line_number: int):
```
weedqian, line 64: Non-critical

format_current_process -> format_current_process_info?

Generally, a process cannot be formated...

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -152,52 +91,82 @@ def __init__(
         self._run_id = run_id
         self._variant_id = variant_id
         self._validate_inputs = validate_inputs
-        multiprocessing_start_method = os.environ.get("PF_BATCH_METHOD")
```
weedqian, line 155: Just a quick question

Did we expose "PF_BATCH_METHOD" to customers in anyways?

Do we need to add a logging here to say "PF_BATCH_METHOD" has been deprecated if we found it's set in ENV?

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -152,52 +91,82 @@ def __init__(
...
-                loaded_tools=flow_executor._loaded_tools,
-                cache_manager=flow_executor._cache_manager,
-            )
+        use_fork = "fork" in sys_start_methods
+        self.context = get_multiprocessing_context("fork" if use_fork else "spawn")
+        self._flow_file = flow_executor._flow_file
+        self._connections = flow_executor._connections
+        self._working_dir = flow_executor._working_dir
+
+        if not use_fork:
```
weedqian, line 101: Just a stupid question. You don't have to modify.

Why we keep the self._executor_creation_func in __init__ func? Can we simply put the logic in __enter__ and remove the self._executor_creation_func?

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -208,65 +177,116 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
 
-    def _timeout_process_wrapper(self, run_start_time: datetime, task_queue: Queue, timeout_time, result_list):
-        healthy_ensured_process = HealthyEnsuredProcess(self._executor_creation_func, self.context)
-        healthy_ensured_process.start_new(task_queue)
+    def _get_process_info(self, index):
+        while True:
+            try:
+                process_id = self._process_info[index].process_id
+                process_name = self._process_info[index].process_name
+                process_info = ProcessInfo(index=index, process_id=process_id, process_name=process_name)
```
weedqian, line 185: Why we recreate the ProcessInfo here?

Hhhilulu: Removed

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: fixed.
Renamed 'format current process info'

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -152,52 +91,82 @@ def __init__(
...
+        self._line_timeout_sec = get_int_env_var("PF_LINE_TIMEOUT_SEC", flow_executor._line_timeout_sec)
         self._output_dir = output_dir
 
     def __enter__(self):
         manager = Manager()
         self._processing_idx = manager.dict()
         self._completed_idx = manager.dict()
 
-        self._inputs_queue = Queue()
+        self._task_queue = Queue()
```
guming-learning, line 130: Why are we using Queue here and manager.Queue in line 135?

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -152,52 +91,82 @@ def __init__(
...
+                self._input_queues,
+                self._output_queues,
+                self._process_info,
+                _process_wrapper,
+                False,
+            )
+
+        self._processes_manager.start_processes()
+
         pool = ThreadPool(self._n_process, initializer=set_context, initargs=(contextvars.copy_context(),))
```
guming-learning, line 170: Maybe rename the `pool` to denote the purpose, maybe something like "monitor_pool" and "self._monitor_pool"?

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -208,65 +177,116 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        result_list.append(result)
+
+    def handle_process_crashed(self, line_number, inputs, run_id, start_time, result_list):
+        bulk_logger.warning(f"Process crashed while executing line {line_number},")
+        ex = ProcessCrashError(line_number)
+        result = self._generate_line_result_for_exception(inputs, run_id, line_number, self._flow_id, start_time, ex)
+        result_list.append(result)
+
+    def _monitor_process_alive(self, process_id):
+        if not psutil.pid_exists(process_id):
```
guming-learning, line 205: This function can be simplified
```suggestion
    return not psutil.pid_exists(process_id)
```

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -208,65 +177,116 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
-            # Handling the timeout of a line execution process.
-            if not completed:
-                bulk_logger.warning(f"Line {line_number} timeout after {timeout_time} seconds.")
-                ex = LineExecutionTimeoutError(line_number, timeout_time)
-                result = self._generate_line_result_for_exception(
-                    inputs, run_id, line_number, self._flow_id, start_time, ex
-                )
-                result_list.append(result)
-                self._completed_idx[line_number] = healthy_ensured_process.format_current_process(line_number, True)
+                # Monitor process aliveness.
```
guming-learning, line 253: No sleep here?

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -208,65 +177,116 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+                    break
+
+                # Handle output queue message.
+                completed = self._handle_output_queue_messages(output_queue, result_list)
+                if completed:
+                    break
+
+            # Handle line execution completed.
+            if completed:
+                self._completed_idx[line_number] = format_current_process(process_name, process_id, line_number)
```
guming-learning, line 265: Seems line 265, 266 are identical with line 276,277 and can be put before the if/else clause?

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: Add logs to prompt users "The 'PF_BATCH_METHOD' has been deprecated"

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: Queue() uses a synchronization primitive called a “lock” to ensure that only one process can access
 the queue at a time
 While manager().Queue() uses a manager object to create a queue that can be shared
between multiple processes.
Use Queue() will case an error: "A SemLock created in a fork context is being shared with a process
in a spawn context. This is not supported".
 So use the manager.Queue().

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: The _handle_output_queue_messages gets the output_queue message within 1 second. I think this is the same with sleep 1s.

src/promptflow/promptflow/executor/_process_manager.py
```diff
@@ -0,0 +1,458 @@
...
+        """
+        if control_signal == ProcessControlSignal.END:
+            self.end_process(i)
+        elif control_signal == ProcessControlSignal.RESTART:
+            self.restart_process(i)
+        elif control_signal == ProcessControlSignal.START:
+            self.new_process(i)
+
+
+def fork_processes_manager(
```
guming-learning, line 378: The name seems confusing, should be something like "create_spawned_fork_process_manager?"

Hhhilulu: fixed

crazygao (dismissed):

src/promptflow/promptflow/executor/flow_executor.py
```diff
@@ -462,6 +461,12 @@ def _exec_batch_with_process_pool(
 
         result_list = []
 
+        # Just for runtime delete the legacy related code
+        # Will removed later.
+        if self._flow_file is None:
```
thy09, line 466: Should raise UnexpectedError with proper message.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -152,121 +91,199 @@ def __init__(
...
+            self._monitor_pool.join()
+
+    def _get_process_info(self, index):
+        while True:
+            try:
+                process_id = self._process_info[index].process_id
+                process_name = self._process_info[index].process_name
+                return (index, process_id, process_name)
+            except KeyError:
+                continue
```
thy09, line 187: Explain the continue here

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -152,121 +91,199 @@ def __init__(
...
-        healthy_ensured_process.start_new(task_queue)
+    def handle_process_crashed(self, line_number, inputs, run_id, start_time, result_list):
+        bulk_logger.warning(f"Process crashed while executing line {line_number},")
+        ex = ProcessCrashError(line_number)
+        result = self._generate_line_result_for_exception(inputs, run_id, line_number, self._flow_id, start_time, ex)
+        result_list.append(result)
 
-        if not healthy_ensured_process.process.is_alive():
-            return
+    def _monitor_process_alive(self, process_id):
```
thy09, line 203: `is_process_alive` might be more straightforward?

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -152,121 +91,199 @@ def __init__(
...
+        index, process_id, process_name = self._get_process_info(index)
 
         while True:
             try:
+                # Get task from task_queue
                 args = task_queue.get(timeout=1)
             except queue.Empty:
-                healthy_ensured_process.end()
+                self._processes_manager.end_process(index)
+                while psutil.pid_exists(process_id):
```
thy09, line 232: Why this? Seems to be an infinite loop if process_id and index doesn't match.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -152,121 +91,199 @@ def __init__(
...
-            if not completed:
-                bulk_logger.warning(f"Line {line_number} timeout after {timeout_time} seconds.")
-                ex = LineExecutionTimeoutError(line_number, timeout_time)
-                result = self._generate_line_result_for_exception(
-                    inputs, run_id, line_number, self._flow_id, start_time, ex
-                )
-                result_list.append(result)
-                self._completed_idx[line_number] = healthy_ensured_process.format_current_process(line_number, True)
+                # Monitor process aliveness.
+                crashed = self._monitor_process_alive(process_id)
```
thy09, line 251: Maybe `crashed = not self._is_process_alive(process_id)`

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -152,121 +91,199 @@ def __init__(
...
+            if completed:
+                self._completed_idx[line_number] = format_current_process_info(process_name, process_id, line_number)
+                log_process_status(process_name, process_id, line_number, is_completed=True)
+            # Handle line execution is not completed.
+            else:
+                # Handle process crashed.
+                if crashed:
+                    self.handle_process_crashed(line_number, inputs, run_id, start_time, result_list)
+                # Handle line execution timeout.
+                else:
```
thy09, line 270: Else means timeout because only crashed or timeouted will go to this code? Maybe add more comments to explain this, or have a signal "timeouted"

src/promptflow/promptflow/executor/flow_executor.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: fixed.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: Add timeouted

Hhhilulu: Add timeouted

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: In fork mode, the main process and the sub spawn process communicate through _process_info. We need to ensure the process has been killed before returning. Otherwise, it may cause the main process have exited but the spawn process is still alive. At this time, a connection error will be reported.

process_id and index are match, process_info uses index as key, process_id and process_name as value.
image

src/promptflow/promptflow/executor/_line_execution_process_pool.py
thy09: It seems to be an infinite loop if it is always not ready, we need to sleep and wait for a while.

Hhhilulu: fixed

thy09 (approved):

lumoslnt (approved):

guming-learning (approved):

weedqian (approved):

crazygao (approved):

PR #1061, opened 2023-11-09
[promptflow][BugFix] Fix output padding when some inputs are missing

zhengfeiwang:
# Description

In some cases, inputs line number may not have full set (e.g. line 1, 2, 4), so that we should not pad the outputs from line number 0. This PR updates the padding function parameter, so that padding will be applied according to the real line number.

# All Promptflow Contribution checklist:
- [x] **The pull request does not introduce [breaking changes].**
- [ ] **CHANGELOG is updated for new features, bug fixes or other significant changes.**
- [x] **I have read the [contribution guidelines](../CONTRIBUTING.md).**
- [ ] **Create an issue and link to the pull request to get dedicated review from promptflow team. Learn more: [suggested workflow](../CONTRIBUTING.md#suggested-workflow).**

## General Guidelines and Best Practices
- [x] Title of the pull request is clear and informative.
- [x] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR. For more information on cleaning up the commits in your PR, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).

### Testing Guidelines
- [x] Pull request includes test coverage for the included changes.

PR #1602, opened 2023-12-26
[promptflow][bugfix] Fix unaligned IO during get details against partial completed Azure run

zhengfeiwang:
# Description

User might run into pandas error, or get unaligned IO during get details against Azure run. The root cause is SDK filter failed lines, so that the line count may be mismatched or unaligned.

This PR targets to fix this by recording inputs line number, sort it and append empty line for those failed lines; to cover this change and guard this behavior, add a pair of "mod N" flow to construct partial completed runs against both data and run.

The fix for local is merged in #1061 .

A screenshot for new get details, which will raise pandas exception before, and is aligned with portal experience:

image

### Testing Guidelines


crazygao:
give one or two get details picture in the Description

src/promptflow/tests/sdk_cli_test/e2etests/test_flow_run.py
```diff
@@ -1080,3 +1080,37 @@ def test_aggregation_node_failed(self, pf):
...
+
+        # batch run against previous run
+        run2 = pf.run(
+            flow=flow_mod3,
+            run=run1,
+            column_mapping={"number": "${run.outputs.output}"},
+        )
+        pf.runs.stream(run2)
+        details2 = pf.get_details(run2)
+        assert len(details2) == 10
```
0mza987, line 1111: Seems the length here becomes 10 which is the valid output length of `run1`. I have a little concern here, since your run2 is not a standard evaluation flow. Let's say your data has another key `groundtruth`, and this run2's column mapping uses both data's ground truth and run1's output, now we know that data has 20 lines, run1 has 20 lines but only 10 is valid, then what's the length of run2's output?

zhengfeiwang: I just modified the second flow and submitted a [run](https://ml.azure.com/prompts/flow/bulkrun/run/5964e473-7d2b-4f9d-a95d-ac7e09b6efcb/details?wsid=/subscriptions/96aede12-2f73-41cb-b983-6d11a904839b/resourceGroups/promptflow/providers/Microsoft.MachineLearningServices/workspaces/promptflow-eastus2euap&tid=72f988bf-86f1-41af-91ab-2d7cd011db47) like you described, there are still 10 lines - which I guess executor did some magic for this.

image

zhengfeiwang: I just take a look on exectuor/batch enginer implementation: in `_merge_input_dicts_by_line`, they will align the data according to the line number, so that the output length should be aligned with the base run (`run1` for your case).

0mza987: Good seems we don't have execution problem here. The only left question is if we should make run2's output length to 10 instead of 20. 

I personally feel it's a bit weird but not a big problem. Maybe ask for @wangchao1230 's opinions.

zhengfeiwang: offline synced with Clement, maybe we can raise this to executor internally.

crazygao (approved):

0mza987 (approved):

wangchao1230 (approved):

src/promptflow/tests/sdk_cli_test/e2etests/test_flow_run.py
```diff
@@ -1080,3 +1080,37 @@ def test_aggregation_node_failed(self, pf):
...
+            data=data_path,
+            column_mapping={"number": "${data.value}"},
+        )
+        pf.runs.stream(run1)
+        details1 = pf.get_details(run1)
+        assert len(details1) == 20
+        assert len(details1.loc[details1["outputs.output"] != "(Failed)"]) == 10
+        # assert to ensure inputs and outputs are aligned
+        for _, row in details1.iterrows():
+            if str(row["outputs.output"]) != "(Failed)":
```
wangchao1230, line 1100: We have a special str(Failed) now. might need to revisit this in future. 
How do user get the failed lines and error message.

zhengfeiwang: agreed, have opened a work item tracking this.

PR #1662, opened 2024-01-04
Use relative path in gen meta/ load tool error message.

huaiyan:
# Description
Use relative path in gen meta/ load tool error message. Absolute paths convey too much extra info.

Move `InvalidSource` from `executor._errors` to `_core._errors` for circular reference.

- [X] **The pull request does not introduce [breaking changes].**
- [X] **I have read the [contribution guidelines](../CONTRIBUTING.md).**

- [X] Title of the pull request is clear and informative.
- [X] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR. For more information on cleaning up the commits in your PR, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).

### Testing Guidelines
- [X] Pull request includes test coverage for the included changes.

### DIFF
--- a/src/promptflow/promptflow/_core/tool_meta_generator.py
+++ b/src/promptflow/promptflow/_core/tool_meta_generator.py
@@ -256,6 +256,7 @@ def generate_python_meta_dict(name, content, source=None):
     return asdict_without_none(generate_python_tool(name, content, source))
 
 
+# Only used in non-code first experience.
 def generate_python_meta(name, content, source=None):
     return json.dumps(generate_python_meta_dict(name, content, source), indent=2)
 
@@ -269,12 +270,12 @@ def generate_tool_meta_dict_by_file(path: str, tool_type: ToolType):
     note that if a python file is passed, correct working directory must be set and should be added to sys.path.
     """
     tool_type = ToolType(tool_type)
-    file = Path(path).resolve()
+    file = Path(path)
     if not file.is_file():
         raise MetaFileNotFound(
             message_format="Generate tool meta failed for {tool_type} tool. Meta file '{file_path}' can not be found.",
             tool_type=tool_type.value,
-            file_path=str(file),
+            file_path=path,  # Use a relative path here to make the error message more readable.
         )
     try:
         content = file.read_text(encoding="utf-8")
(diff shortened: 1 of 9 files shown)


src/promptflow/promptflow/_core/tool_meta_generator.py
```diff
@@ -269,12 +270,12 @@ def generate_tool_meta_dict_by_file(path: str, tool_type: ToolType):
     note that if a python file is passed, correct working directory must be set and should be added to sys.path.
     """
     tool_type = ToolType(tool_type)
-    file = Path(path).resolve()
+    file = Path(path).resolve(strict=False)
```
thy09, line 273: Why change strict=False?

huaiyan: The original issue was Path(path).resolve() raised `FileNotFoundError`.
Which will only happen when `strict=True`.
Although didn't find any python version's default strict value is True, manually set it as False to avoid future issue.

thy09: We may add comments here to explain this.

huaiyan: After re-consider the implementation, the `resolve` should not be necessary there.
Just remove it.

src/promptflow/promptflow/_core/tool_meta_generator.py
```diff
@@ -256,6 +256,7 @@ def generate_python_meta_dict(name, content, source=None):
     return asdict_without_none(generate_python_tool(name, content, source))
 
 
+# Only used in non-code first experience.
```
thy09, line 259: Do we still have this?

huaiyan: Still has reference from runtime code
https://msdata.visualstudio.com/Vienna/_git/PromptFlow?path=/src/promptflow-sdk/promptflow/runtime/app.py&version=GBdevs/robbenwang_flow_mapping&line=762&lineEnd=762&lineStartColumn=13&lineEndColumn=57&lineStyle=plain&_a=contents

thy09 (approved):

lumoslnt (approved):

PR #1698, opened 2024-01-10
[Executor] Support node_concurrency when running an async flow

Jasmin3q:
# Description

Use asyncio.semaphore to support concurrency control.

When node_concurrency is set to 2:
sync_passthrough1 and async_passthrough1 can run concurrently:

When node_concurrency is set to 1:
Only one of sync_passthrough1 and async_passthrough1 can at the same time:

### DIFF
--- a/src/promptflow/promptflow/executor/_async_nodes_scheduler.py
+++ b/src/promptflow/promptflow/executor/_async_nodes_scheduler.py
@@ -29,7 +29,6 @@ class AsyncNodesScheduler:
         node_concurrency: int,
     ) -> None:
         self._tools_manager = tools_manager
-        # TODO: Add concurrency control in execution
         self._node_concurrency = node_concurrency
 
     async def execute(
@@ -46,7 +45,8 @@ class AsyncNodesScheduler:
             flow_logger.info(
                 "Current thread is not main thread, skip signal handler registration in AsyncNodesScheduler."
             )
-
+        loop = asyncio.get_running_loop()
+        self._semaphore = asyncio.Semaphore(self._node_concurrency, loop=loop)
         parent_context = contextvars.copy_context()
         executor = ThreadPoolExecutor(
             max_workers=self._node_concurrency, initializer=set_context, initargs=(parent_context,)
@@ -104,6 +104,10 @@ class AsyncNodesScheduler:
             self._create_node_task(node, dag_manager, context, executor): node for node in dag_manager.pop_ready_nodes()
         }
 
+    async def run_task_with_semaphore(self, coroutine):
+        async with self._semaphore:
+            return await coroutine
+
     def _create_node_task(
         self,
         node: Node,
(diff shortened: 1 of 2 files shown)


guming-learning (approved):

src/promptflow/promptflow/executor/_async_nodes_scheduler.py
```diff
@@ -114,7 +118,8 @@ def _create_node_task(
         f = self._tools_manager.get_tool(node.name)
         kwargs = dag_manager.get_node_valid_inputs(node, f)
         if inspect.iscoroutinefunction(f):
-            task = context.invoke_tool_async(node, f, kwargs)
+            coroutine = context.invoke_tool_async(node, f, kwargs)
+            task = self.run_task_with_semaphore(coroutine)
```
thy09, line 122: Seems only async task is under the concurrency limit. Sync/async tasks should share the same limit.

Jasmin3q: Added run_task_with_semaphore to sync tasks.

guming-learning (approved):

src/promptflow/promptflow/executor/_async_nodes_scheduler.py
```diff
@@ -119,7 +123,7 @@ def _create_node_task(
             task = self._sync_function_to_async_task(executor, context, node, f, kwargs)
         # Set the name of the task to the node name for debugging purpose
         # It does not need to be unique by design.
-        return asyncio.create_task(task, name=node.name)
+        return asyncio.create_task(self.run_task_with_semaphore(task), name=node.name)
```
thy09, line 126: It seems that the task is created before we put it to the semaphore, we may add more detailed comments why this works:
1 For async task, does it mean that the the task will not be executed before calling `create_task`?
2 For sync task, does it mean that even though the task is put to the threading pool, thread.start will only be triggered after `create_task`? 

It would be better to confirm these details and add more comments here.

src/promptflow/tests/executor/e2etests/test_async.py
```diff
@@ -0,0 +1,41 @@
...
+                    node_times[api_call["node_name"]]["end"] = api_call["end_time"]
+            return node_times
+
+        # When node_concurrency is 1, nodes execute one by one
+        flow_result = executor.exec_line({"input_str": "Hello"}, node_concurrency=1)
+        node_times = get_node_times(flow_result.run_info.api_calls, nodes)
+        assert node_times[nodes[0]]["end"] < node_times[nodes[1]]["start"]
+        assert node_times[nodes[1]]["end"] < node_times[nodes[2]]["start"]
+
+        # When node_concurrency is 2, two nodes execute concurrently
```
thy09, line 37: How do we know this concurrency is 2 not 3 or more? The test logic is a little hard to understand, maybe add more comments.

src/promptflow/promptflow/executor/_async_nodes_scheduler.py
Jasmin3q: After test, we can confirm that the task will not run until calling create_task. Added comments explaining the calling sequence in _create_node_task.

src/promptflow/tests/executor/e2etests/test_async.py
Jasmin3q: In the new test case,  we count the max_node_concurrency during the flow execution according to the flow_result. At last we ensure it is expected and will not exceed the concurrency level we set.

thy09 (approved):

guming-learning (approved):

PR #1757, opened 2024-01-16
fix get user identity info

Stephen1993:
# Description
image

### Testing Guidelines


src/promptflow/promptflow/_sdk/operations/_run_operations.py
```diff
@@ -233,6 +233,8 @@ def delete(
         name: str,
     ) -> None:
         """Delete run permanently.
+        Caution: This operation will delete the run permanently from your local disk.
```
zhengfeiwang, line 236: seems irelevant to this PR?

src/promptflow/promptflow/azure/_restclient/flow_service_caller.py
```diff
@@ -134,7 +134,7 @@ def _get_user_identity_info(self):
         import jwt
 
         token = get_arm_token(credential=self._credential)
-        decoded_token = jwt.decode(token.token, options={"verify_signature": False})
+        decoded_token = jwt.decode(token, options={"verify_signature": False})
```
zhengfeiwang, line 137: consider adding a UT for this function `_get_user_identity_info`? so that we can avoid regression in the future on same case

src/promptflow/promptflow/_sdk/operations/_run_operations.py
Stephen1993: Yes, this is a modification that was missed in the previous PR, I will change it in this PR.

src/promptflow/promptflow/azure/_restclient/flow_service_caller.py
D-W-: Do we have test for `pf.flows.test`? maybe we can add one if not

D-W-, line 137: Return the decoded token in `get_arm_token` to avoid decode again here?

Stephen1993: Temporarily not modified, the reason is:
1. Make the return value and semantics of the "**get_arm_token**" function clearer. Although there is an internal decode, it is only to verify the correctness of the token, and the parameters used during the decode are inconsistent with here
2. Decode currently performs very well, repeating it once has no significant impact.

D-W-: add a UT for this func then?

D-W-: Maybe add some annotation on function's interface to mark the return value changed to string instead of token object?

Stephen1993: done, modified the flow.list test, which will cover this function.

Stephen1993: ok, the return type has been added.

Stephen1993: We will not add test for now, as adding the following two fields to the token will cause the existing test replay mode to fail.
image

Stephen1993: image

src/promptflow/tests/test_configs/recordings/test_flow_operations_TestFlow_test_list_flows.yaml
```diff
@@ -209,4 +209,446 @@ interactions:
     status:
       code: 200
       message: OK
-version: 1
\ No newline at end of file
+- request:
```
zhengfeiwang, line 212: revert this file change

zhengfeiwang (changes requested):

src/promptflow/promptflow/azure/_restclient/flow_service_caller.py
wangchao1230: shall we create task to track this?

wangchao1230 (dismissed):

zhengfeiwang (approved):

D-W- (approved):

PR #1791, opened 2024-01-19
[Internal][Executor] Fix tests to be compatible with various versions of python and os

PeiwenGaoMS:
# Description

Fix tests to be compatible with various versions of python and os.

### Testing Guidelines

### DIFF
--- a/src/promptflow/tests/executor/e2etests/test_batch_timeout.py
+++ b/src/promptflow/tests/executor/e2etests/test_batch_timeout.py
@@ -23,6 +23,9 @@ class TestBatchTimeout:
     def setup_method(self):
         BatchEngine.register_executor(FlowLanguage.Python, MockPythonExecutorProxy)
 
+    def teardown_method(self):
+        BatchEngine.register_executor(FlowLanguage.Python, PythonExecutorProxy)
+
     @pytest.mark.parametrize(
         "flow_folder",
         [
--- a/src/promptflow/tests/test_configs/flows/flow_with_langchain_traces/test_langchain_traces.py
+++ b/src/promptflow/tests/test_configs/flows/flow_with_langchain_traces/test_langchain_traces.py
@@ -1,12 +1,10 @@
 import os
 
-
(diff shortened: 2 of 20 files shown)


src/promptflow/tests/test_configs/flows/flow_with_langchain_traces/test_langchain_traces.py
```diff
@@ -34,5 +32,7 @@ def test_langchain_traces(question: str, conn: AzureOpenAIConnection):
         content=question
     )
 
-    return agent.run(message)
-
+    try:
+        return agent.run(message)
+    except Exception as e:
+        return str(e)
```
guming-learning, line 38: Why swallow the  exception here?

src/promptflow/tests/executor/e2etests/test_batch_timeout.py
```diff
@@ -23,6 +23,9 @@ class TestBatchTimeout:
     def setup_method(self):
         BatchEngine.register_executor(FlowLanguage.Python, MockPythonExecutorProxy)
 
+    def teardown_method(self):
```
guming-learning, line 26: Where is this function called?

PeiwenGaoMS: This method will be called after all tests for this class have completed. This is the usage of pytest framework.
https://docs.pytest.org/en/7.1.x/how-to/xunit_setup.html

src/promptflow/tests/test_configs/flows/flow_with_langchain_traces/test_langchain_traces.py
lumoslnt: We may encounter output parsing error when run langchain agent coincidentally, basically the error is because the randomness of the output of LLM, when it did not strictly follow the instruction. So we need to add try-catch here to avoid blocking tests.

zzn2 (approved):

thy09 (approved):

PR #1839, opened 2024-01-24
[PromptFlow][ProcessPool] Solve spawned fork process manager hangs for a long time when parameters are passed incorrectly.

Hhhilulu:
# Description

**Problem**
Currently, in fork mode, when spawn process raises an exception, the process will hang and not raise the exception to main process. So we need to solve the problem.

**Solution**

1. In fork mode, the spawn process is the bridge between the main and fork processes. If the spawned process is no longer running, exit the main process.
2. Add try-catch in the spawned process, if an exception occurred, log the error message.
3. In the Monitor thread, if not get the process info within the 60s, raise an exception.

### Testing Guidelines

### DIFF
--- a/src/promptflow/promptflow/executor/_line_execution_process_pool.py
+++ b/src/promptflow/promptflow/executor/_line_execution_process_pool.py
@@ -457,6 +457,7 @@ class LineExecutionProcessPool:
             while time.time() - ensure_spawn_process_healthy_start_time < 6:
                 if psutil.Process(self._managed_process_id).status() == "zombie":
                     return
+                time.sleep(1)
 
         with RepeatLogTimer(
             interval_seconds=self._log_interval,
--- a/src/promptflow/promptflow/executor/_line_execution_process_pool.py
+++ b/src/promptflow/promptflow/executor/_line_execution_process_pool.py
@@ -36,7 +36,6 @@ from promptflow.executor._errors import (
     ProcessCrashError,
     ProcessInfoObtainedTimeout,
     ProcessTerminatedTimeout,
-    SpawnedForkProcessManagerStartFailure,
 )
 from promptflow.executor._process_manager import ForkProcessManager, SpawnProcessManager
 from promptflow.executor._result import LineResult
@@ -171,19 +170,15 @@ class LineExecutionProcessPool:
                 self._flow_create_kwargs,
                 **common_kwargs,
             )
-            self._processes_manager.start_processes()
-            # If the spawned process that created the fork process has not started successfully,
-            # raise 'SpawnedForkProcessManagerStartFailure' exception and without continue execution.
-            if not self._processes_manager._is_spawned_fork_process_manager_healthy:
-                ex = SpawnedForkProcessManagerStartFailure()
-                raise ex
         else:
             executor_creation_func = partial(FlowExecutor.create, **self._flow_create_kwargs)
             # 1. Create input_queue, output_queue, and _process_info in the main process.
             # 2. Spawn _n_process sub-process and pass the above queue/dict to these sub-process to transfer information
             # between main process and sub process.
             self._processes_manager = SpawnProcessManager(executor_creation_func, **common_kwargs)
-            self._processes_manager.start_processes()
+
+        self._processes_manager.start_processes()
+        self._processes_manager.ensure_healthy()
 
         monitor_pool = ThreadPool(self._n_process, initializer=set_context, initargs=(contextvars.copy_context(),))
         self._monitor_pool = monitor_pool
--- a/src/promptflow/promptflow/executor/_line_execution_process_pool.py
+++ b/src/promptflow/promptflow/executor/_line_execution_process_pool.py
@@ -205,8 +205,6 @@ class LineExecutionProcessPool:
                 # try again.
                 time.sleep(1)
                 continue
-            except queue.Empty:
-                pass
             except Exception as e:
                 raise Exception(f"Unexpected error occurred while get process info. Exception: {e}")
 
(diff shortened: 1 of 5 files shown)


src/promptflow/promptflow/executor/_process_manager.py
```diff
@@ -225,6 +226,7 @@ def start_processes(self):
             ),
         )
         process.start()
+        return process.pid
```
guming-learning, line 229: Manager process id should not be returned here, because it's not consistent with the interface of SpawnProcessManager anymore.
Consider set `self.spawned_process_manager_pid`, then the caller can get it by fork_process_manager.spawned_process_manager_pid

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -446,6 +450,15 @@ def run(self, batch_inputs):
         result_list = []
         run_start_time = datetime.utcnow()
 
+        # In fork mode, the spawn process is the bridge between the main and fork processes.
+        # If the spawned process is no longer running, exit the main proccess.
+        if self._use_fork:
+            ensure_spawn_process_healthy_start_time = time.time()
+            while time.time() - ensure_spawn_process_healthy_start_time < 6:
+                if psutil.Process(self._managed_process_id).status() == "zombie":
+                    return
```
guming-learning, line 459: Please add some log here.

Hhhilulu: Added

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -446,6 +451,21 @@ def run(self, batch_inputs):
         result_list = []
         run_start_time = datetime.utcnow()
 
+        # In fork mode, the spawned process is the bridge between the main and fork processes.
+        # If the spawned process is no longer running, exit the main proccess.
+        if self._use_fork:
```
thy09, line 456: The idea looks good to me.
As the implmenetation, I prefer the proposal like this:
`self._process_manager.ensure_healthy()` then all related logic could be moved to the implementation of the process manager instead of `if else` in the process_pool logic.

guming-learning (approved):

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -170,14 +171,19 @@ def __enter__(self):
...
+                raise ex
         else:
             executor_creation_func = partial(FlowExecutor.create, **self._flow_create_kwargs)
             # 1. Create input_queue, output_queue, and _process_info in the main process.
             # 2. Spawn _n_process sub-process and pass the above queue/dict to these sub-process to transfer information
             # between main process and sub process.
             self._processes_manager = SpawnProcessManager(executor_creation_func, **common_kwargs)
-
-        self._processes_manager.start_processes()
+            self._processes_manager.start_processes()
```
thy09, line 186: I think we can have a method in the base class `ensure_healthy()`, the the logic here is common:
```
self._process_manager.start_processes()
self._process_manager.ensure_healthy()
```

Then then we can check and raise the exception in the process manager implementation.
And we can also have some validation logic for spawn mode.

guming-learning (approved):

guming-learning (dismissed):

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -193,6 +194,7 @@ def _get_process_info(self, index):
         start_time = time.time()
         while True:
             try:
+                self._processes_manager.ensure_healthy()
```
thy09, line 197: Put `ensure_healthy` outside try? So we don't need to specially handle the PromptflowException

src/promptflow/promptflow/executor/_process_manager.py
```diff
@@ -368,6 +399,13 @@ def handle_signals(self, control_signal, i):
             self.new_process(i)
 
 
+def create_spawned_fork_process_manager_wrapper(*args, **kwargs):
+    try:
+        create_spawned_fork_process_manager(*args, **kwargs)
+    except Exception as e:
```
thy09, line 405: Do we still need this?

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: If we put ensure_healthy outside, it will return true and end after 1 minute.

src/promptflow/promptflow/executor/_process_manager.py
Hhhilulu: Yes, if we remove the wrapper, errors of the incorrect parameters will not generate a corresponding log to prompt the developer, but will hang for 1minutes until get process info fails.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
thy09: I mean something like this:
```
while True:
    self._process_manager.ensure_healthy()
    try:
        xxxx
    except
        xxxx
```

thy09 (approved):

src/promptflow/promptflow/executor/_line_execution_process_pool.py
Hhhilulu: fixed

src/promptflow/promptflow/executor/_process_manager.py
Hhhilulu: remove

guming-learning (approved):

PR #1862, opened 2024-01-26
[Internal][Executor] Add batch error dict to batch result and update it if batch run timeout

PeiwenGaoMS:
# Description

Add batch error dict to batch result and update it if batch run timeout.

### Testing Guidelines


src/promptflow/promptflow/batch/_batch_engine.py
```diff
@@ -262,16 +263,16 @@ async def _exec(
...
+            results, is_timeout = self._executor_proxy._exec_batch(
+                batch_inputs,
+                output_dir,
+                run_id,
+                batch_timeout_sec=self._batch_timeout_sec,
+                line_timeout_sec=self._line_timeout_sec,
             )
+            line_results.extend(results)
         else:
             await self._exec_batch(line_results, batch_inputs, run_id)
```
thy09, line 277: Add TODO about batch_timeout for this code path,

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -330,6 +340,11 @@ def _monitor_workers_and_process_tasks_in_thread(
 
             self._processing_idx.pop(line_number)
 
+        # If the task queue is not empty, it indicates that the exit from the while loop was due to a batch run timeout.
+        # In this case, we should set is_timeout to True.
+        if not task_queue.empty():
```
thy09, line 345: This logic is tricky, why in this case `is_timeout` is not set.

The logic here highly relies on the assumption of the implementation somewhere else, this is not a recommended practice.

src/promptflow/promptflow/batch/_batch_engine.py
PeiwenGaoMS: Added.

src/promptflow/promptflow/executor/_line_execution_process_pool.py
PeiwenGaoMS: Updated.

thy09 (approved):

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -349,6 +360,10 @@ def _monitor_workers_and_process_tasks_in_thread(
 
             self._processing_idx.pop(line_number)
 
+        # If the while loop exits due to batch run timeout, we should set is_timeout to True.
+        if self._batch_timeout_expired(batch_start_time):
```
Hhhilulu, line 364: It seems that if `self._is_timeout` is set to `True `in the monitor thread, it will be re-evaluated here. Additionally, if a batch run fails due to a batch timeout, it appears that this code will be executed regardless of whether task_queue is empty. Perhaps we can write it like this:
`self._is_timeout = self._is_timeout or self._batch_timeout_expired(batch_start_time)`

PeiwenGaoMS: Updated.

Hhhilulu (approved):

thy09 (approved):

PR #2001, opened 2024-02-07
[promptflow][internal] Support batch run scenario when generate line run

zhengfeiwang:
# Description

This PR targets to support batch run scenario when generate line run object.

**Change list**

- Collect spans considering both test and batch run scenario
- Set up exporter in line process, so that traces in those processes can be exported
- [BugFix] Missing trace name in some cases, commit from @brynn-code 
- [BugFix] Copy context when in notebook, otherwise operation context will get lost
- [BugFIx] When set referenced batch run id, make sure we are setting the string value
- [Workaround] Let line process wait for 10 seconds to ensure spans are flushed, this will be refined later by @thy09 

The remaining knwon issue is for line process, it might be killed before the exported finished, so we might miss some root span for some lines.

### Testing Guidelines

### DIFF
--- a/src/promptflow/promptflow/executor/_line_execution_process_pool.py
+++ b/src/promptflow/promptflow/executor/_line_execution_process_pool.py
@@ -367,7 +367,10 @@ class LineExecutionProcessPool:
         # wait 10 seconds to ensure the spans are exported before the process is killed
         # otherwise there will be some missing spans for lines
         # TODO (Task 2950080): make line process wait for spans flush
-        time.sleep(10)
+        from promptflow._trace._start_trace import _is_tracer_provider_configured
+
+        if _is_tracer_provider_configured():
+            time.sleep(10)
 
         # End the process when the batch timeout is exceeded or when all lines have been executed.
         self._processes_manager.end_process(index)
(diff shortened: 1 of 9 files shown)


wangchao1230 (approved):

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -364,6 +364,11 @@ def _monitor_workers_and_process_tasks_in_thread(
         # If the while loop exits due to batch run timeout, we should set is_timeout to True if we didn't set it before.
         self._is_timeout = self._is_timeout or self._batch_timeout_expired(batch_start_time)
 
+        # wait 10 seconds to ensure the spans are exported before the process is killed
+        # otherwise there will be some missing spans for lines
+        # TODO (Task 2950080): make line process wait for spans flush
+        time.sleep(10)
```
zhengfeiwang, line 370: MARK & highlight: this change might lead to check-enforcer timeout/CI break, as we make every line 10 seconds longer

src/promptflow/promptflow/executor/_line_execution_process_pool.py
```diff
@@ -364,6 +364,11 @@ def _monitor_workers_and_process_tasks_in_thread(
         # If the while loop exits due to batch run timeout, we should set is_timeout to True if we didn't set it before.
         self._is_timeout = self._is_timeout or self._batch_timeout_expired(batch_start_time)
 
+        # wait 10 seconds to ensure the spans are exported before the process is killed
```
wangchao1230, line 367: executor should call some flush command before it exit

src/promptflow/promptflow/executor/_line_execution_process_pool.py
zhengfeiwang: added a check before the sleep, however, I believe all batch run from SDK should internally set up the exporter, so the check shall not work better.

thy09 (approved):

wangchao1230 (approved):

PR #1823, opened 2024-01-23
[promptflow][internal] Integrate JS visualize bundle

zhengfeiwang:
# Description

For internal usage, integrate JavaScript visualize bundle to repository.

**Change List**:

- Update code to dump JavaScript bundle from package to where the HTML locates
- Remove code that uses CDN link
- Update Compliance Check YAML to ad-hoc delete this JavaScript bundle, otherwise the pipeline will fail.

The new HTML looks like:

image

# All Promptflow Contribution checklist:
- [x] **The pull request does not introduce [breaking changes].**
- [ ] **CHANGELOG is updated for new features, bug fixes or other significant changes.**
- [x] **I have read the [contribution guidelines](../CONTRIBUTING.md).**
- [ ] **Create an issue and link to the pull request to get dedicated review from promptflow team. Learn more: [suggested workflow](../CONTRIBUTING.md#suggested-workflow).**

## General Guidelines and Best Practices
- [x] Title of the pull request is clear and informative.
- [x] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR. For more information on cleaning up the commits in your PR, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).

### Testing Guidelines
- [ ] Pull request includes test coverage for the included changes.

PR #2057, opened 2024-02-20
[promptflow][internal] Fix visualize HTML missing JS bug

zhengfeiwang:
# Description

#1823 integrates visualize JS bundle into our repo, and `pf run visualize` will internally dump that JS bundle next to the HTML file. However, in some scenarios (e.g., C#), visualize page is generated via local PFS visualize API, where we have no good place to pass that JS bundle; meanwhile, we might **stop updating run visualization**, so that we might not need to update the JS bundle anymore. So the solution is back to the CDN link, which is currently seen as static - so directly write in HTML template.

### Testing Guidelines


src/promptflow/promptflow/_sdk/data/visualize.j2
```diff
@@ -10,6 +10,6 @@
     <script>
       window.bulk_test_details_data = {{ data }}
     </script>
-    <script src="{{ js_path }}"></script>
+    <script src="https://sdk-bulk-test-endpoint.azureedge.net/bulk-test-details/view/0.0.35/bulkTestDetails.min.js"></script>
```
zhengfeiwang, line 13: I understand as this CDN link will not be updated anymore, so we don't need the suffix `?version=1` anymore.

+ @RealKai42 to help confirm

wangchao1230 (approved):

RealKai42 (approved):

thy09 (approved):

PR #2200, opened 2024-03-04
[internal][bugfix] Create a placeholder for still running line runs

zhengfeiwang:
# Description

We will ignore those still-running line runs, but when the user runs a time-consuming flow, he cannot see anything until this line run terminated - this is very annoyed. So in this PR, we will create a placeholder for such case, so that user can happily view something, even before the line(s) terminated.

**Running traces**

image

**Trace UI url change**

The new url will look like `http://localhost:<port>/v1.0/ui/traces/?#run=<run>`, there will be a **#** in the path, which is the way that React/UX route, similar to what we do with `?` in backend. It's not easy to avoid this breaking change, and as 1) we can update the print url 2) tracing is still a private preview feature, so this change shall be acceptable.

### Testing Guidelines


src/promptflow/promptflow/_constants.py
```diff
@@ -64,6 +64,7 @@ class AvailableIDE:
 # trace related
 OTEL_RESOURCE_SERVICE_NAME = "promptflow"
 DEFAULT_SPAN_TYPE = "default"
+RUNNING_LINE_RUN_STATUS = "Running"
```
zhengfeiwang, line 67: may wait for UX changes in this PR, otherwise, users will hit the "No span data" issue.

src/promptflow/promptflow/_sdk/operations/_trace_operations.py
```diff
@@ -34,99 +37,98 @@ def list_line_runs(
...
+        orm_spans_group_by_trace_id = ORMLineRun.list_with_runs(runs)
+        # ORM entities to SDK entities
+        spans_group_by_trace_id = list()
+        for orm_spans in orm_spans_group_by_trace_id:
+            spans_group_by_trace_id.append([Span._from_orm_object(orm_span) for orm_span in orm_spans])
+        # aggregation logic is different when runs are specified
+        # so will not call `_aggregate_spans` here
+        grouped_spans = {run: defaultdict(list) for run in runs}
+        for trace_id_spans in spans_group_by_trace_id:
+            obo_span: Span = trace_id_spans[0]
```
wangchao1230, line 112: what's obo stands for?

zhengfeiwang: On Behalf Of - the attributes in the list are same, so just pick the first to on behalf of them, there are some comments explaining this within this file, maybe I can short copy some here.

src/promptflow/promptflow/_sdk/_service/static/index.html
```diff
@@ -3,11 +3,12 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <link rel="icon" href="{{ url_for('static', filename='favicon.ico') }}" />
-    <title>Promptflow traces</title>
+    <link rel="icon" href="/v1.0/ui/traces/assets/favicon.ico" />
+    <title>Trace View</title>
+    <script type="module" crossorigin src="/v1.0/ui/traces/assets/index-4cTGdARK.js"></script>
```
crazygao, line 8: crossorigin?

zhengfeiwang: not sure, these files will be maintained by UX devs; any comment on this? maybe I need to engage UX dev on this

src/promptflow/promptflow/_sdk/_tracing.py
```diff
@@ -86,13 +86,13 @@ def _print_tracing_url_from_local(
     exp: typing.Optional[str] = None,
     run: typing.Optional[str] = None,
 ) -> None:
-    url = f"http://localhost:{pfs_port}/v1.0/ui/traces"
+    url = f"http://localhost:{pfs_port}/v1.0/ui/traces/"
     if run is not None:
-        url += f"?run={run}"
+        url += f"?#run={run}"
```
0mza987, line 91: why need this extra #?

zhengfeiwang: this should be the way how UX/React route

zhengfeiwang: double confirmed with Kaiyi, you can regard this style with `#` is the way how front end manage the route, so we might need this breaking change.

0mza987: good to know, can  add the explanation to pr description.

0mza987 (approved):

src/promptflow/promptflow/_constants.py
```diff
@@ -68,6 +68,7 @@ class AvailableIDE:
...
```
0mza987, line 71: Another minor change, the status `Ok` is strange, can make it to `OK`? May do this in next pr tho.
image

0mza987: And why is it `OK` instead of `Completed` which is more aligned with our run status in the first place?

zhengfeiwang: ah, this is something that open telemetry defines: <https://opentelemetry.io/docs/specs/otel/trace/api/#set-status>

image

i don't like it, but, that is the standard...

wangchao1230 (approved):

PR #2230, opened 2024-03-06
[internal][tracing] Introduce environment variable `PF_TRACING_SKIP_LOCAL_SETUP` to skip tracing local setup

zhengfeiwang:
# Description

This PR targets to introduce a new environment variable to skip tracing local setup. This can be set in runtime, so that we won't invoke PFS and setup local OTLP collector endpoint when in cloud runtime environment.

This PR also adds the CI workflow against tracing unit test.

### Testing Guidelines


src/promptflow/promptflow/tracing/_start_trace.py
```diff
@@ -82,9 +78,48 @@ def start_trace(*, session: typing.Optional[str] = None, **kwargs):
     else:
         operation_context._add_otel_attributes(SpanAttributeFieldName.REFERENCED_LINE_RUN_ID, ref_line_run_id)
 
-    # trace provider from pf config, for local to cloud scenario
+    if _skip_tracing_local_setup():
```
wangchao1230, line 81: I thought the config here should be some port?
So we can avoid a bool setting?

zhengfeiwang: why port? runtime will have their own logic to setup environment variables, so we don't need to execute the following lines that only works for local.

wangchao1230: A port that points the Trace endpoint, this setting can also be used in serving scenario

zhengfeiwang: executor will always call `setup_exporter_from_environ`, this will consume info from environment variable; so I prefer let the runtime take control for those environs - let's offline sync on this

wangchao1230 (approved):

riddlexu (approved):

Issue #2276, opened 2024-03-08
[BUG] broken flow_file when use default working_dir in Flow.from_yaml()


mrgomdev:
**Describe the bug**
When flow_file only specified, working_dir get set as flow_file.parent. However, working_dir / flow_file is not the valid path when flow_file is relative path of more than 2 depths. For example:
flow_file='./Flow-hello/flow.dag.yaml'
working_dir=None
working_dir='./Flow-hello/'.absolute()
working_dir / flow_file = './Flow-hello/Flow-hello/flow.dag.yaml'

**How To Reproduce the bug**
Steps to reproduce the behavior, how frequent can you experience the bug:
1. flow_file is relative path of more than 2 depths. For example, './Flow-hello/flow.dag.yaml'
2. working_dir is None
3. Flow.from_yaml(flow_file=flow_file, working_dir=working_dir)

**Expected behavior**
No error on Flow.from_yaml(...)

**Screenshots**

**Running Information(please complete the following information):**
 - Promptflow Package Version using `pf -v`: [e.g. 0.0.102309906]
 - Operating System: [e.g. Ubuntu 20.04, Windows 11]
 - Python Version using `python --version`: [e.g. python==3.10.12]

**Additional context**

PR #2277, opened 2024-03-08
fix: broken flow_file when use default working_dir

mrgomdev:
From issue #2276 

# Description

### Testing Guidelines

### DIFF
--- a/src/promptflow/promptflow/contracts/flow.py
+++ b/src/promptflow/promptflow/contracts/flow.py
@@ -687,8 +687,9 @@ class Flow(FlowBase):
     def _parse_working_dir(cls, flow_file: Path, working_dir=None) -> Path:
         if working_dir is None:
             working_dir = Path(flow_file).resolve().parent
+            flow_file = Path(flow_file).name
         working_dir = Path(working_dir).absolute()
-        return working_dir
+        return flow_file, working_dir
 
     @classmethod
     def _update_working_dir(cls, working_dir: Path):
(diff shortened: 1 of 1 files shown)


mrgomdev:
Oh, it looks not covering all usages of _parse_working_dir(). This PR is not valid.

Better solution is to prevent current beahvior in from_yaml(), like [the case in flow_executor.py](https://github.com/microsoft/promptflow/blob/c9165c32b7569f415f9b730e1046253f19481c05/src/promptflow/promptflow/executor/flow_executor.py#L282).

src/promptflow/promptflow/contracts/flow.py
```diff
@@ -697,7 +698,7 @@ def _update_working_dir(cls, working_dir: Path):
     @classmethod
     def from_yaml(cls, flow_file: Path, working_dir=None) -> "Flow":
         """Load flow from yaml file."""
-        working_dir = cls._parse_working_dir(flow_file, working_dir)
+        flow_file, working_dir = cls._parse_working_dir(flow_file, working_dir)
```
thy09, line 701: I guess we can have `flow_file = flow_file.resolve()` when working_dir is not None

mrgomdev: It would work! You mean, right before original L#701 `with open(working_dir / flow_file, (...), flow_file should be resolved when working_dir is not None, right?

D-W-, line 701: Could you add a UT for the fixed scenario?

PR #2354, opened 2024-03-14
fix sdk system error

Stephen1993:
# Description

image

### Testing Guidelines


src/promptflow/promptflow/exceptions.py
```diff
@@ -54,12 +54,14 @@ def __init__(
         message_format="",
         target: ErrorTarget = ErrorTarget.UNKNOWN,
         module=None,
+        privacy_info: List[str] = None,
```
wangchao1230, line 57: add this to docstring?

src/promptflow/promptflow/_cli/_utils.py
```diff
@@ -389,11 +389,14 @@ def get_secret_input(prompt, mask="*"):
     - Ignore control characters and print warning message.
     """
     if not isinstance(prompt, str):
-        raise TypeError(f"prompt must be a str, not ${type(prompt).__name__}")
+        e = TypeError(f"prompt must be a str, not ${type(prompt).__name__}")
+        raise UserErrorException(str(e), privacy_info=[]) from e
```
wangchao1230, line 393: why empty privacy_info here?

wangchao1230: I remember you want to use None to determine the behavior here. But it looks strange for devs not knowing that.

wangchao1230: Let's use error format for this case?

Stephen1993: Because I want to record messages, I need **privacy_info**  is not **None**.
image

Stephen1993: I will change to using **message_format**, but it still feels strange because **message parameters** are not needed.
image

src/promptflow/promptflow/exceptions.py
Stephen1993: done

src/promptflow/promptflow/_utils/load_data.py
```diff
@@ -59,12 +59,13 @@ def _bfs_dir(dir_path: List[str]) -> Tuple[List[str], List[str]]:
...
     files, dirs = [], []
     for path in dir_path:
-        for filename in os.listdir(path):
-            file = Path(path, filename).resolve()
-            if file.is_file():
-                files.append(str(file))
-            else:
-                dirs.append(str(file))
+        if os.path.exists(path):
+            for filename in os.listdir(path):  # Path does not exist and will throw FileNotFoundError
```
crazygao, line 63: I think we should try catch and throw this as UserError?
If we suppress this message we lose some information about the folder?

Stephen1993: Based on the context message, the purpose of this function is to search for valid files and directories, so I prefer to add a check to see if the directory is valid instead of throwing an error.

src/promptflow/promptflow/_cli/_utils.py
```diff
@@ -389,11 +389,14 @@ def get_secret_input(prompt, mask="*"):
...
+        e = TypeError(f"prompt must be a str, not ${type(prompt).__name__}")
+        raise UserErrorException(message_format=str(e)) from e
     if not isinstance(mask, str):
-        raise TypeError(f"mask argument must be a one-character str, not ${type(mask).__name__}")
+        e = TypeError(f"mask argument must be a one-character str, not ${type(mask).__name__}")
+        raise UserErrorException(message_format=str(e)) from e
     if len(mask) != 1:
-        raise ValueError("mask argument must be a one-character str")
+        e = ValueError("mask argument must be a one-character str")
+        raise UserErrorException(message_format=str(e)) from e
```
0mza987, line 399: why do we need the inner `ValueError`, feel like we can simply change the original `TypeError` to `UserErrorException`?

src/promptflow/promptflow/_sdk/_load_functions.py
```diff
@@ -63,7 +64,7 @@ def load_common(
             **kwargs,
         )
     except Exception as e:
-        raise Exception(f"Load entity error: {e}") from e
+        raise UserErrorException(f"Load entity error: {e}", privacy_info=[str(e)]) from e
```
0mza987, line 67: Think we can create a `LoadFunctionError` that inherits from UserErrorException and use it in all errors inside this py file?

src/promptflow/promptflow/_cli/_utils.py
Stephen1993: Because it can display the most primitive error types:
image

src/promptflow/promptflow/_sdk/_load_functions.py
Stephen1993: I think it might be better to keep the most primitive error type because it is an unknown error and we hope the dashboard will display the specific error type in the end.

crazygao (approved):

wangchao1230 (approved):

0mza987 (approved):

PR #2433, opened 2024-03-21
[Internal] Support Union connections for register_apis

dorisjoy:
Issue, currently the register_apis only support below api which only have one connection type:
image

But we have a requirement to change the connection to Union to include ServerlessConnection like below:
image

So here make change to let the register_apis support Union connections.

Add UT for it.

### DIFF
--- a/src/promptflow/promptflow/_core/tools_manager.py
+++ b/src/promptflow/promptflow/_core/tools_manager.py
@@ -10,7 +10,7 @@ import traceback
 import types
 from functools import partial
 from pathlib import Path
-from typing import Callable, Dict, List, Mapping, Optional, Tuple, Union
+from typing import Callable, Dict, List, Mapping, Optional, Tuple, Union, get_origin, get_args
 
 from promptflow._core._errors import (
     InputTypeMismatch,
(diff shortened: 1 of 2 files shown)


thy09 (approved):

src/promptflow/promptflow/_core/tools_manager.py
```diff
@@ -474,14 +486,17 @@ def _register(provider_cls, collection, type):
...
-        if annotation_type_name in connections:
-            api_name = provider_cls.__name__
-            module_logger.debug(f"Add connection type {annotation_type_name} to api {api_name} mapping")
-            connection_type_to_api_mapping[annotation_type_name] = api_name
+        types = get_all_supported_types(param.annotation)
+        for type in types:
+            if type in connections:
+                module_logger.debug(f"Add connection type {type} to api {api_name} mapping")
+                connection_type_to_api_mapping[type] = api_name
+                should_break = True
```
chjinche, line 498: should_break is used but not assigned? May lead to `NameError` in python

dorisjoy: Fixed

chjinche (approved):

thy09 (approved):

PR #2479, opened 2024-03-26
[Internal] Upload the events in span to Blob.

riddlexu:
# Description

In this PR, we upload the events in span to blob.
- Init the BlobClient.
  - The azureml uri has pattern "azureml://subscriptions/$sub/resourcegroups/$rg/workspaces/$ws/datastores/$storename/paths/.promptflow/.trace/$collection_id/$trace_id/$span_id/$id"
  - We cannot get the datastore name from blob client. Blob client only has container name, so we need to get the base uri including datastore name in collector, and pass it to span client.
  - Get default datastore may take 2~5s. So, we add a cache.
- Upload event to blob and add the azureml uri to "external_event_data_uris".
- Clear the "attributes" in event to reduce the span size.

Test case: [query link](https://int.ml.azure.com/prompts/trace/list?wsid=%2Fsubscriptions%2F96aede12-2f73-41cb-b983-6d11a904839b%2FresourceGroups%2Fpromptflow%2Fproviders%2FMicrosoft.MachineLearningServices%2Fworkspaces%2Fpromptflow-eastus-dev&searchText=%7B%22batchRunId%22%3A%22chat_with_pdf_variant_0_20240328_204721_774665%22%7D&tid=72f988bf-86f1-41af-91ab-2d7cd011db47)

image

Remainder: To upload data to blob, user need to have the role "Storage Blob Data Contributor" to the storage account of workspace.


src/promptflow/promptflow/azure/_storage/cosmosdb/span.py
```diff
@@ -42,21 +47,40 @@ def __init__(self, span: SpanEntity, collection_id: str, created_by: dict) -> No
...
         try:
-            return client.create_item(body=self.to_dict())
+            return cosmos_client.create_item(body=self.to_dict())
         except CosmosResourceExistsError:
             return None
 
     def to_dict(self) -> Dict[str, Any]:
         return {k: v for k, v in self.__dict__.items() if v}
+
+    def _persist_events(self, blob_container_client: ContainerClient, blob_base_uri: str):
```
liucheng-ms, line 73: Shall we add slash check for `blob_base_uri` in the beginning?
```python
if not blob_base_uri.endswith('/'):
    blob_base_uri += '/'
```

src/promptflow/promptflow/azure/_storage/cosmosdb/span.py
```diff
@@ -42,21 +47,40 @@ def __init__(self, span: SpanEntity, collection_id: str, created_by: dict) -> No
...
+        idx = 0
+        for event in self.events:
```
liucheng-ms, line 75: May be we can use enumerate for idx
```python
for idx, event in enumerate(self.events):
```

src/promptflow/promptflow/azure/_storage/cosmosdb/span.py
```diff
@@ -42,21 +47,40 @@ def __init__(self, span: SpanEntity, collection_id: str, created_by: dict) -> No
...
+            blob_client = blob_container_client.get_blob_client(self._event_path(idx))
+            blob_client.upload_blob(event_data)
+
+            event[SpanEventFieldName.ATTRIBUTES] = {}
+            self.external_event_data_uris.append(f"{blob_base_uri}{self._event_path(idx)}")
+
+            idx += 1
+
+    def _event_path(self, idx: int) -> str:
+        return f".promptflow/.trace/{self.collection_id}/{self.context[SpanContextFieldName.TRACE_ID]}/{self.id}/{idx}"
```
liucheng-ms, line 86: How about extract the hard code part as constant for maintainability.

```python
EVENT_PATH_PREFIX = ".promptflow/.trace/"

def _event_path(self, idx: int) -> str:
    return f"{EVENT_PATH_PREFIX}{self.collection_id}/{self.context[SpanContextFieldName.TRACE_ID]}/{self.id}/{idx}"
```

src/promptflow/promptflow/azure/_storage/cosmosdb/span.py
```diff
@@ -42,21 +47,40 @@ def __init__(self, span: SpanEntity, collection_id: str, created_by: dict) -> No
...
 
     def to_dict(self) -> Dict[str, Any]:
         return {k: v for k, v in self.__dict__.items() if v}
+
+    def _persist_events(self, blob_container_client: ContainerClient, blob_base_uri: str):
+        idx = 0
+        for event in self.events:
+            event_data = json.dumps(event)
+            blob_client = blob_container_client.get_blob_client(self._event_path(idx))
+            blob_client.upload_blob(event_data)
```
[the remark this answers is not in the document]
liucheng-ms: Shall we retry?

src/promptflow/promptflow/azure/_storage/cosmosdb/span.py
riddlexu: Added. Thanks.

src/promptflow/promptflow/azure/_storage/cosmosdb/span.py
riddlexu: Fixed. Thanks.

src/promptflow/promptflow/azure/_storage/cosmosdb/span.py
riddlexu: Fixed. Thanks.

src/promptflow/promptflow/_sdk/_service/apis/collector.py
```diff
@@ -181,3 +187,56 @@ def _try_write_trace_to_cosmosdb(
         stack_trace = traceback.format_exc()
         logger.error(f"Failed to write trace to cosmosdb: {e}, stack trace is {stack_trace}")
         return
+
+
+def _try_get_blob_container_client(
```
huaiyan, line 192: What's the latency for getting blob container client?
Do we need to move it in worker thread?

src/promptflow/promptflow/_sdk/_service/apis/collector.py
```diff
@@ -181,3 +187,56 @@ def _try_write_trace_to_cosmosdb(
...
+        # Datastore is a notion of AzureML, it is not a notion of Blob Storage.
+        # So, we cannot get datastore name by blob client.
+        # To generate the azureml uri has datastore name, we need to generate the uri here and pass in to db client.
+
+        return ContainerClient(
+            account_url=account_url, container_name=default_datastore.container_name, credential=credential
+        ), LONG_URI_FORMAT.format(subscription_id, resource_group_name, workspace_name, default_datastore.name, "")
+
+    except Exception as e:
+        stack_trace = traceback.format_exc()
```
huaiyan, line 240: If fail to get the client, should we fail the request directly?

src/promptflow/promptflow/_sdk/_service/apis/collector.py
riddlexu: Nice catch. Get default datastore may take 2~5s. We may have a cache for blob client.

src/promptflow/promptflow/azure/_storage/cosmosdb/span.py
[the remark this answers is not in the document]
riddlexu: For retry, blob client has default retry policy([doc link](https://learn.microsoft.com/en-us/python/api/overview/azure/storage-blob-readme?view=azure-python#retry-policy-configuration))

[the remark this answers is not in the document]
riddlexu: And offline synced, we should fail for such case.

src/promptflow/promptflow/_sdk/_service/apis/collector.py
riddlexu: Fixed. Will raise the exception.

zhengfeiwang (dismissed):

src/promptflow-azure/promptflow/azure/_storage/blob/client.py
```diff
@@ -0,0 +1,109 @@
...
+
+from promptflow.exceptions import UserErrorException
+
+datastore_map = {}
+client_map = {}
+_thread_lock = threading.Lock()
+_cache_timeout = 60 * 5  # Will try to refresh default datastore if exceed 5 minutes
+
+
+def get_client(
```
liucheng-ms, line 22: Should we specify the client type more clearly? It appears that by default, this function returns a `ContainerClient` for the default blob container.

Perhaps we could rename it to `get_datastore_container_client` for clarity?

src/promptflow-azure/promptflow/azure/_storage/blob/client.py
```diff
@@ -0,0 +1,109 @@
...
+            credential = DefaultAzureCredential()
+
+        default_datastore = get_default_datastore(subscription_id, resource_group_name, workspace_name, credential)
+
+        storage_endpoint = _get_storage_endpoint_from_metadata()
+        account_url = STORAGE_ACCOUNT_URLS[DatastoreType.AZURE_BLOB].format(
+            default_datastore.account_name, storage_endpoint
+        )
+
+        from azure.ai.ml.constants._common import LONG_URI_FORMAT
```
liucheng-ms, line 43: Is there a specific rationale for importing at the function level? I generally prefer imports at the top of the file. :)

src/promptflow-azure/promptflow/azure/_storage/blob/client.py
```diff
@@ -0,0 +1,109 @@
...
+
+        # Datastore is a notion of AzureML, it is not a notion of Blob Storage.
+        # So, we cannot get datastore name by blob client.
+        # To generate the azureml uri has datastore name, we need to generate the uri here and pass in to db client.
+        return ContainerClient(
```
liucheng-ms, line 48: I find the tuple a bit complex to understand at a glance. Could we consider simplifying it as shown below?

```python
        container_client = ContainerClient(  
            account_url=account_url, container_name=default_datastore.container_name, credential=credential  
        )  
        azureml_uri = LONG_URI_FORMAT.format(subscription_id, resource_group_name, workspace_name, default_datastore.name, "")  
  
        return container_client, azureml_uri
```

src/promptflow-azure/promptflow/azure/_storage/blob/client.py
```diff
@@ -0,0 +1,109 @@
...
+
+    datastore_key = _get_datastore_client_key(subscription_id, resource_group_name, workspace_name)
+    datastore = _get_datastore_from_map(datastore_key=datastore_key)
+    if datastore is None:
+        with _thread_lock:
+            datastore = _get_datastore_from_map(datastore_key=datastore_key)
+            if datastore is None:
+                datastore = _get_default_datastore(subscription_id, resource_group_name, workspace_name, credential)
+                datastore_map[datastore_key] = {
+                    "expire_at": datetime.datetime.now() + datetime.timedelta(0, _cache_timeout),
```
liucheng-ms, line 70: Could we consider using `datetime.timedelta(seconds=_cache_timeout)` directly to simplify the code?

src/promptflow-azure/promptflow/azure/_storage/blob/client.py
```diff
@@ -0,0 +1,109 @@
...
+                    "datastore": datastore,
+                }
+    return datastore
+
+
+def _get_datastore_from_map(datastore_key: str):
```
liucheng-ms, line 76: Seems it is retrieving from cache. Shall we use name `_get_datastore_from_cache`.

src/promptflow-azure/promptflow/azure/_storage/blob/client.py
```diff
@@ -0,0 +1,109 @@
...
+    datastore = datastore_map.get(datastore_key, None)
```
liucheng-ms, line 77: Default is `None`. We can just use `datastore = datastore_map.get(datastore_key)`

src/promptflow-azure/promptflow/azure/_storage/cosmosdb/span.py
```diff
@@ -42,21 +46,43 @@ def __init__(self, span: SpanEntity, collection_id: str, created_by: dict) -> No
...
         return {k: v for k, v in self.__dict__.items() if v}
+
+    def _persist_events(self, blob_container_client: ContainerClient, blob_base_uri: str):
+        if not blob_base_uri.endswith("/"):
+            blob_base_uri += "/"
+
+        for idx, event in enumerate(self.events):
+            event_data = json.dumps(event)
+            blob_client = blob_container_client.get_blob_client(self._event_path(idx))
+            blob_client.upload_blob(event_data)
```
liucheng-ms, line 79: Shall we handle error for upload_blob?

src/promptflow-azure/promptflow/azure/_storage/cosmosdb/span.py
```diff
@@ -42,21 +46,43 @@ def __init__(self, span: SpanEntity, collection_id: str, created_by: dict) -> No
...
             return
 
+        if self.events and blob_container_client is not None and blob_base_uri is not None:
+            self._persist_events(blob_container_client, blob_base_uri)
+
         resource_attributes = self.resource.get(SpanFieldName.ATTRIBUTES, None)
         if resource_attributes is None:
             return
 
         from azure.cosmos.exceptions import CosmosResourceExistsError
```
liucheng-ms, line 62: Shall we move this to top?

src/promptflow-azure/promptflow/azure/_storage/cosmosdb/span.py
```diff
@@ -42,21 +46,43 @@ def __init__(self, span: SpanEntity, collection_id: str, created_by: dict) -> No
...
 
         try:
-            return client.create_item(body=self.to_dict())
+            return cosmos_client.create_item(body=self.to_dict())
         except CosmosResourceExistsError:
             return None
```
liucheng-ms, line 67: Shall we align return style with line 60

src/promptflow-devkit/promptflow/_sdk/_service/apis/collector.py
```diff
@@ -134,18 +135,33 @@ def _try_write_trace_to_cosmosdb(
...
+
         # Load created_by info first time may slow. So, we load it in parallel for warm up.
         created_by_thread = ThreadWithContextVars(target=get_created_by_info_with_cache)
         created_by_thread.start()
 
-        get_client(CosmosDBContainerName.LINE_SUMMARY, subscription_id, resource_group_name, workspace_name, credential)
+        # Get default blob may be slow. So, we have a cache for default datastore.
+        from promptflow.azure._storage.blob.client import get_client as get_blob_client
+
+        blob_container_client, blob_base_uri = get_blob_client(
```
liucheng-ms, line 151: I know there is cache for container client. So we do not use thread here? What's the purpose of using thread for cosmos cilent but not using thread for container client?

src/promptflow-azure/promptflow/azure/_storage/blob/client.py
riddlexu: Fixed. Thanks.

src/promptflow-azure/promptflow/azure/_storage/blob/client.py
riddlexu: Fixed. Thanks.

src/promptflow-azure/promptflow/azure/_storage/blob/client.py
riddlexu: Fixed. Thanks.

src/promptflow-azure/promptflow/azure/_storage/blob/client.py
riddlexu: Fixed. Thanks.

src/promptflow-azure/promptflow/azure/_storage/blob/client.py
riddlexu: Fixed. Thanks.

src/promptflow-azure/promptflow/azure/_storage/blob/client.py
riddlexu: Fixed. Thanks.

src/promptflow-azure/promptflow/azure/_storage/blob/client.py
```diff
@@ -0,0 +1,109 @@
...
+    if datastore is None:
```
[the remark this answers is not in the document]
riddlexu: Fixed. Thanks.

src/promptflow-azure/promptflow/azure/_storage/cosmosdb/span.py
riddlexu: Fixed. Thanks.

src/promptflow-devkit/promptflow/_sdk/_service/apis/collector.py
riddlexu: The logic of get blob client run at main thread, not need to start a new thread.

src/promptflow-azure/promptflow/azure/_storage/cosmosdb/span.py
riddlexu: I think we should not? We don't need to upload the data to blob when we do not persist span at cosmos db.

src/promptflow-azure/promptflow/azure/_storage/cosmosdb/span.py
riddlexu: I think we should throw exception here? Because we don't know how to handle blob storage fail currently.

src/promptflow-azure/promptflow/azure/_storage/blob/client.py
```diff
@@ -0,0 +1,109 @@
...
+                    "expire_at": datetime.datetime.now() + datetime.timedelta(seconds=_cache_timeout),
+                    "datastore": datastore,
+                }
+    return datastore
+
+
+def _get_datastore_from_cache(datastore_key: str):
+    datastore = _datastore_cache.get(datastore_key)
+
+    if datastore and datastore["expire_at"] > datetime.datetime.now():
```
huaiyan, line 81: Just out of curiosity, do we have plan to reuse the cache logic in future?

src/promptflow-azure/promptflow/azure/_storage/cosmosdb/span.py
```diff
@@ -42,21 +46,43 @@ def __init__(self, span: SpanEntity, collection_id: str, created_by: dict) -> No
...
-    def persist(self, client):
+    def persist(self, cosmos_client: ContainerProxy, blob_container_client: ContainerClient, blob_base_uri: str):
         if self.id is None or self.partition_key is None or self.resource is None:
             return
 
         resource_attributes = self.resource.get(SpanFieldName.ATTRIBUTES, None)
         if resource_attributes is None:
             return
 
+        if self.events and blob_container_client is not None and blob_base_uri is not None:
```
huaiyan, line 59: Not None check is not necessary for now, right?

src/promptflow-azure/promptflow/azure/_storage/cosmosdb/span.py
```diff
@@ -42,21 +46,43 @@ def __init__(self, span: SpanEntity, collection_id: str, created_by: dict) -> No
...
+            return cosmos_client.create_item(body=self.to_dict())
         except CosmosResourceExistsError:
-            return None
+            return
 
     def to_dict(self) -> Dict[str, Any]:
         return {k: v for k, v in self.__dict__.items() if v}
+
+    def _persist_events(self, blob_container_client: ContainerClient, blob_base_uri: str):
+        if not blob_base_uri.endswith("/"):
```
huaiyan, line 73: Put this compatibility logic to client.py?

src/promptflow-azure/promptflow/azure/_storage/cosmosdb/span.py
```diff
@@ -42,21 +46,43 @@ def __init__(self, span: SpanEntity, collection_id: str, created_by: dict) -> No
...
+            blob_client.upload_blob(event_data)
+
+            event[SpanEventFieldName.ATTRIBUTES] = {}
+            self.external_event_data_uris.append(f"{blob_base_uri}{self._event_path(idx)}")
+
+    EVENT_PATH_PREFIX = ".promptflow/.trace"
+
+    def _event_path(self, idx: int) -> str:
+        trace_id = self.context[SpanContextFieldName.TRACE_ID]
+        return f"{self.EVENT_PATH_PREFIX}/{self.collection_id}/{trace_id}/{self.id}/{idx}"
```
huaiyan, line 88: Maybe use partition key instead of collection id?
Or we could pass in batch run id as collection id.
Do we plan to have collection id concept for batch run?

src/promptflow-azure/promptflow/azure/_storage/blob/client.py
```diff
@@ -0,0 +1,109 @@
...
+
+    datastore_key = _get_datastore_client_key(subscription_id, resource_group_name, workspace_name)
+    datastore = _get_datastore_from_cache(datastore_key=datastore_key)
+    if datastore is None:
+        with _thread_lock:
+            datastore = _get_datastore_from_cache(datastore_key=datastore_key)
+            if datastore is None:
+                datastore = _get_default_datastore(subscription_id, resource_group_name, workspace_name, credential)
+                _datastore_cache[datastore_key] = {
+                    "expire_at": datetime.datetime.now() + datetime.timedelta(seconds=_cache_timeout),
```
huaiyan, line 72: How do we get the timeout threshold?

src/promptflow-azure/promptflow/azure/_storage/cosmosdb/span.py
riddlexu: Yes. The check is just for defend..

src/promptflow-azure/promptflow/azure/_storage/cosmosdb/span.py
riddlexu: Maybe we should not use partition key because use meaningful field is better? For the runtime, I will pass batch_run_id as collection_id.

src/promptflow-azure/promptflow/azure/_storage/blob/client.py
riddlexu: Yes. We should create something like cache_with_ttl. Seems the python std lib has no such lib.

src/promptflow-azure/promptflow/azure/_storage/cosmosdb/span.py
riddlexu: Fixed. Thanks.

src/promptflow-azure/promptflow/azure/_storage/blob/client.py
riddlexu: We set the same timeout with cosmos db client. When cosmos db cache expire, the blob also expire. This will reduce the cosmos db timeout impact.

src/promptflow-azure/promptflow/azure/_storage/cosmosdb/span.py
huaiyan: If so, do we expect let UX query by collection_id or batch_run_id?

riddlexu: Yes.

src/promptflow-azure/promptflow/azure/_storage/cosmosdb/span.py
liucheng-ms: OK

src/promptflow-azure/promptflow/azure/_storage/cosmosdb/span.py
liucheng-ms: Sorry for confusion. I intended to say the import line. Shall we align the style of importing?

src/promptflow-devkit/promptflow/_sdk/_service/apis/collector.py
```diff
@@ -134,18 +135,33 @@ def _try_write_trace_to_cosmosdb(
...
+        )
+        line_summary_client_thread.start()
+
         # Load created_by info first time may slow. So, we load it in parallel for warm up.
         created_by_thread = ThreadWithContextVars(target=get_created_by_info_with_cache)
         created_by_thread.start()
 
-        get_client(CosmosDBContainerName.LINE_SUMMARY, subscription_id, resource_group_name, workspace_name, credential)
+        # Get default blob may be slow. So, we have a cache for default datastore.
+        from promptflow.azure._storage.blob.client import get_datastore_container_client
```
liucheng-ms, line 149: Is there any particular reason to import inside a function? It’s more of a style question, and the decision is entirely up to you. 😆

src/promptflow-devkit/promptflow/_sdk/_service/apis/collector.py
```diff
@@ -134,18 +135,33 @@ def _try_write_trace_to_cosmosdb(
...
+
+        blob_container_client, blob_base_uri = get_datastore_container_client(
```
liucheng-ms, line 151: Curious on the reason why blob client does not use thread, As I saw existed cosmos db client is wrapped by `ThreadWithContextVars`

src/promptflow-devkit/promptflow/_sdk/_service/apis/collector.py
riddlexu: If not enable local to cloud trace, we should not import azure related lib here.

liucheng-ms: Which means `_try_write_trace_to_cosmosdb` should not be called at all?

riddlexu: Yes. But for current implementation, we need to check the span data to decide if the cloud trace is enable. And we may not want such logic appear in upper level handle.

liucheng-ms (approved):

huaiyan (approved):

zhengfeiwang (approved):

PR #2537, opened 2024-03-29
Support ServerlessConnection in Embedding

dorisjoy:
Support ServerlessConnection in Embedding


src/promptflow-tools/promptflow/tools/embedding.py
```diff
@@ -10,6 +10,14 @@
 from promptflow.connections import AzureOpenAIConnection, OpenAIConnection
 
 
+def is_serverless_connection(connection):
```
chjinche, line 13: Use this compatible process? Then you can keep using `isinstance`.
try:
    from promptflow.connections import ServerlessConnection
except ImportError:
    # If unable to import ServerlessConnection, define a placeholder class to allow isinstance checks to pass.
    # ServerlessConnection was introduced in pf version 1.6.0.
    class ServerlessConnection:
        pass

src/promptflow-tools/promptflow/tools/embedding.py
```diff
@@ -33,6 +45,13 @@ def embedding(connection: Union[AzureOpenAIConnection, OpenAIConnection], input:
             input=input,
             model=model
         ).data[0].embedding
+    elif is_serverless_connection(connection):
```
chjinche, line 48: do we plan to add ut in this pr?

melionel: Yes, we should. I want to add the ut after I tested in local. However I didn't get an embedding deployment for test now.

src/promptflow-tools/promptflow/tools/embedding.py
```diff
@@ -18,8 +25,12 @@ class EmbeddingModel(str, Enum):
 
 @tool
 @handle_openai_error()
-def embedding(connection: Union[AzureOpenAIConnection, OpenAIConnection], input: str, deployment_name: str = "",
-              model: EmbeddingModel = EmbeddingModel.TEXT_EMBEDDING_ADA_002):
+def embedding(
+    connection: Union[AzureOpenAIConnection, OpenAIConnection],
```
DaweiCai, line 29: Union[AzureOpenAIConnection, OpenAIConnection, ServerlessConnection] ？

melionel: good catch

DaweiCai (approved):

chjinche (approved):

PR #2640, opened 2024-04-03
Append v1 to ServerlessConnection in _tool_resolver.py

dorisjoy:
Append v1 to ServerlessConnection in _tool_resolver.py

### DIFF
--- a/src/promptflow-core/promptflow/executor/_tool_resolver.py
+++ b/src/promptflow-core/promptflow/executor/_tool_resolver.py
@@ -469,8 +469,8 @@ class ToolResolver:
                 # 2. Register_apis doesn't support Union;
                 # So in order to release this append "/v1" function without breaking old promptflow package.
                 # We move this logic here.
-                # This is a short-term solution. After MaaS supports OpenAI compatible rest API, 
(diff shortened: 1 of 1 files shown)


weedqian (dismissed):

zzn2 (dismissed):

src/promptflow-core/promptflow/executor/_tool_resolver.py
```diff
@@ -462,9 +462,21 @@ def _resolve_llm_connection_with_provider(connection):
             # This is a fallback for the case that ServerlessConnection related tool is not ready
             # in legacy versions, then we can directly use OpenAIConnection.
             if isinstance(connection, ServerlessConnection):
+                # The ServerlessConnection should be passed into promptflow-tools as it is.
```
DaweiCai, line 465: Also add this comment:  This is a short-term solution. After MaaS supports OpenAI compatible rest API, we should remove the V1 tricky here.

dorisjoy: added

zzn2 (approved):

Bowen-Guo (approved):

PR #2775, opened 2024-04-12
feat: support pf.run for dynamic callable

elliotzh:
# Description

Support `pf.run` for dynamic callable. If the callable entry is dynamic, we won't support resume for it.

### Testing Guidelines

### DIFF
--- a/src/promptflow-devkit/promptflow/_sdk/_orchestrator/run_submitter.py
+++ b/src/promptflow-devkit/promptflow/_sdk/_orchestrator/run_submitter.py
@@ -153,7 +153,7 @@ class RunSubmitter:
         )
         try:
             batch_engine = BatchEngine(
-                flow.path,
+                run._dynamic_callable or flow.path,
                 flow.code,
                 connections=connections,
                 entry=flow.entry if isinstance(flow, FlexFlow) else None,
--- a/src/promptflow-devkit/promptflow/_sdk/entities/_run.py
+++ b/src/promptflow-devkit/promptflow/_sdk/entities/_run.py
@@ -196,6 +196,9 @@ class Run(YAMLTranslatableMixin):
         self._identity = kwargs.get("identity", {})
         self._outputs = kwargs.get("outputs", None)
         self._command = kwargs.get("command", None)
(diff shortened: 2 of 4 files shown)


src/promptflow-devkit/promptflow/_sdk/entities/_run.py
```diff
@@ -196,6 +196,9 @@ def __init__(
         self._identity = kwargs.get("identity", {})
         self._outputs = kwargs.get("outputs", None)
         self._command = kwargs.get("command", None)
+        # To support `pf.run` for dynamic callable, we need to create a run whose target function is a dynamic callable.
+        # TODO: such run is not resumable, not sure if we need specific error message for this case.
```
wangchao1230, line 200: I think we can add a properties for this actually

elliotzh: You mean make it public? or add it to `__init__`? Here my concern, this property will be lost when we do `pf.runs.get`, so it will be strange if we make it a public attribute - it's hard for customer to figure out when this property shouldn't be `None`.

elliotzh: And I don't think it's a good idea to add this argument to `__init__` if we don't give it a public `attribute`. Or you mean we can make it a private attribute first?

src/promptflow-devkit/promptflow/_sdk/_orchestrator/run_submitter.py
```diff
@@ -153,7 +153,7 @@ def _submit_bulk_run(
         )
         try:
             batch_engine = BatchEngine(
-                flow.path,
+                run._dynamic_callable or flow.path,
```
crazygao, line 156: so dynamic callable will drop the path?
all features related to paths will be out of usage?

wangchao1230 (approved):

src/promptflow-devkit/promptflow/_sdk/_orchestrator/run_submitter.py
elliotzh: it won't. it will use the class address as entry and generate a temp `flow.flex.yaml` for path related features.

crazygao (approved):

PR #2786, opened 2024-04-13
Add possibility to select logging level.

nick863:
# Description

This PR address the issue from the work item 3096911. Essentially, we want to control the logging verbosity. I have made the changes in the logging and added unit test.
To better test changes I have also added mypy configuration to pyproject.toml and  fixed all the encountered issues.

### Testing Guidelines


src/promptflow-evals/tests/unittests/test_f1_evaluator.py
```diff
@@ -0,0 +1,36 @@
...
+'''
+import pytest
+import logging
+
+from unittest.mock import patch
+
+from promptflow.evals.evaluators.f1_score import F1ScoreEvaluator
+
+
+class TestF1ScoreEvaluator:
```
ninghu, line 14: Just a suggestion: We probably don't need a unit-test file for each evaluator.  If the intension here is only to test the logging, would suggest to rename the class to `TestEvaluatorLogging` or something similar. Same to the test file name.

src/promptflow-evals/pyproject.toml
```diff
@@ -95,3 +95,12 @@ name = "Contract forbidden modules"
 type = "forbidden"
 source_modules = ["promptflow.evals"]
 forbidden_modules = []
+
+[tool.mypy]
```
ninghu, line 99: Thank you for enabling the mypy check!

src/promptflow-core/promptflow/core/_serving/flow_invoker.py
```diff
@@ -59,7 +60,10 @@ def __init__(
         init_kwargs: dict = None,
         **kwargs,
     ):
-        self.logger = kwargs.get("logger", LoggerFactory.get_logger("flowinvoker"))
+        self.logger = kwargs.get(
+            "logger",
+            LoggerFactory.get_logger("flowinvoker",
+                                     verbosity=kwargs.get('log_level') or logging.INFO))
```
ninghu, line 66: Is it expected to use `logging.INFO` if `log_level` is not provided? Would it override the log_level that was set when getting the logger instance? Wonder whether `get_pf_logging_level()` should be used here instead.

src/promptflow-evals/promptflow/evals/evaluators/fluency/__init__.py
```diff
@@ -4,20 +4,25 @@
...
+
 from promptflow.client import load_flow
 from promptflow.entities import AzureOpenAIConnection
 from pathlib import Path
 
 
 class FluencyEvaluator:
-    def __init__(self, model_config: AzureOpenAIConnection, deployment_name: str):
+    def __init__(self, model_config: AzureOpenAIConnection, deployment_name: str,
+                 log_level: Optional[int] = None):
```
ninghu, line 16: Is this `log_level` requirement only for evalualtors? Have we considered overriding the log level inside evalualtors?

src/promptflow-evals/tests/unittests/test_f1_evaluator.py
nick863: Fixed.

src/promptflow-evals/promptflow/evals/evaluators/fluency/__init__.py
nick863: According to the task: https://msdata.visualstudio.com/Vienna/_workitems/edit/3096911/ we want to avoid evaluator to print the Promptflow logs, or at least make it configurable.

ninghu (dismissed):

src/promptflow-core/promptflow/core/_serving/flow_invoker.py
nick863: Fixed, switched to using `get_pf_logging_level()`

nagkumar91 (dismissed):

ninghu (dismissed):

src/promptflow-evals/promptflow/evals/evaluators/coherence/__init__.py
```diff
@@ -12,12 +14,15 @@
 
 
 class CoherenceEvaluator:
-    def __init__(self, model_config: AzureOpenAIModelConfiguration):
+    def __init__(self, model_config: AzureOpenAIModelConfiguration,
+                 log_level: Optional[int] = None):
```
wangchao1230, line 18: a int parameter for customer is not a good idea. What's the possible value here?

src/promptflow-devkit/promptflow/_sdk/entities/_flows/base.py
```diff
@@ -145,6 +146,7 @@ def __init__(
         **kwargs,
     ):
         self.variant = kwargs.pop("variant", None) or {}
+        self._log_level = kwargs.pop("log_level", None)
```
wangchao1230, line 149: I think this setting should not be in Flow obj as it's changes static loggers which is not part of flow

src/promptflow-evals/promptflow/evals/evaluators/content_safety/hate_unfairness.py
```diff
@@ -30,7 +33,7 @@ def __init__(self, project_scope: dict, credential: Optional[TokenCredential] =
         # Load the flow as function
         current_dir = Path(__file__).resolve().parent
         flow_dir = current_dir / "flow"
-        self._flow = load_flow(source=flow_dir)
+        self._flow = load_flow(source=flow_dir, log_level=log_level)
```
wangchao1230, line 36: could we directly set the logger here?

wangchao1230: Or just give user some doc that they can change logging behavior by setting some env variable.

nick863: The main problem here is that all the log records, the [task](https://msdata.visualstudio.com/Vienna/_workitems/edit/3096911/) are not coming from flow itself, it is from invoker/executor infrastructure.
We are using two mechanisms of getting loggers. If user will set the `PF_LOGGING_LEVEL`, the flow executor logs will ignore it, because the environment variable will have to be set before importing promptflow, as the logger is initialized during import (see the `logger_utils.py`).
```python
flow_logger = get_logger("execution.flow")
```

Another problem is that we initialize the "flowinvoker" logger on flow invoker level, using different mechanism, which will respect `PF_LOGGING_LEVEL` (the code is given before this PR):
```python
self.logger = kwargs.get("logger", LoggerFactory.get_logger("flowinvoker"))
```
So that the logger created/set on the flow level will not have the effect. All

PR #2889, opened 2024-04-19
Handle job cancellation for root span

huaiyan:
# Description
1. Record KeyboardInterrupt exception for span (otlp's code doesn't handle BaseException but only Exception)
2. Raise KeyboardInterrupt in FlowNodesScheduler
3. Remove `with ThreadPoolExecutor` to guarantee KeyboardInterrupt could be received by outside.

Why make this change: **we want to support cancellation for span.**

We have 2 schedulers, one is FlowNodesScheduler, another is AsyncNodesScheduler.

For most cases, we are using FlowNodesScheduler.
In FlowNodesScheduler, there is `with ThreadPoolExecutor`, which will wait all running threads finish, no matter what exception is raised. That's why our cancellation is blocked.
image

So, in this PR, remove the `with ThreadPoolExecutor` to make the KeyboardInterrupt could be received by outside.

Why also change other tests:
For now, our exception could be received by outside immediately, the exception type and execution time are affected.

For test_batch_timeout, situation is very complex.
The test set 5 seconds batch run timeout and 600 seconds line run timeout.
But actually, for batch run execution, line run timeout is calculated by batch run timeout. So, 600 seconds won't take effect, the real line run timeout is less than 5 seconds.
But before this change, even we raise line run timeout exception, we can't exit ThreadPoolExecutor to finish line run.
From process perspective, the line run has never finish, so assign one `BatchExecutionTimeoutError`.
After this change, line run timeout exception could be raised out soon, so the behavior is changed.

### Testing Guidelines

### DIFF
--- a/src/promptflow-core/promptflow/executor/_flow_nodes_scheduler.py
+++ b/src/promptflow-core/promptflow/executor/_flow_nodes_scheduler.py
@@ -115,7 +115,11 @@ class FlowNodesScheduler:
                 if timeout_task and timeout_task.done():
                     raise LineExecutionTimeoutError(self._context._line_number, line_timeout_sec)
                 self._execute_nodes(executor)
+            event.set()
+            executor.shutdown()
         except Exception as e:
+            # When meet exception, mark event to exit timeout task.
+            event.set()
             err_msg = "Flow execution has failed."
             if isinstance(e, LineExecutionTimeoutError):
                 err_msg = f"Line execution timeout after {line_timeout_sec} seconds."
@@ -127,9 +131,6 @@ class FlowNodesScheduler:
                 unfinished_future.cancel()
             # Even we raise exception here, still need to wait all running jobs finish to exit.
             raise e
-        finally:
-            # Cancel timeout task no matter the execution is finished or failed.
-            event.set()
         for node in self._dag_manager.bypassed_nodes:
             self._dag_manager.completed_nodes_outputs[node] = None
         return self._dag_manager.completed_nodes_outputs, self._dag_manager.bypassed_nodes
--- a/src/promptflow-core/promptflow/executor/_flow_nodes_scheduler.py
+++ b/src/promptflow-core/promptflow/executor/_flow_nodes_scheduler.py
@@ -119,8 +119,6 @@ class FlowNodesScheduler:
             event.set()
             executor.shutdown()
         except Exception as e:
-            # When meet exception, mark event to exit timeout task.
-            event.set()
             err_msg = "Flow execution has failed."
             if isinstance(e, LineExecutionTimeoutError):
                 err_msg = f"Line execution timeout after {line_timeout_sec} seconds."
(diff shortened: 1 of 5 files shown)


src/promptflow-core/promptflow/executor/flow_executor.py
```diff
@@ -847,7 +847,7 @@ def _exec_inner_with_trace(
     ):
         # need to get everytime to ensure tracer is latest
         otel_tracer = otel_trace.get_tracer("promptflow")
-        with otel_tracer.start_as_current_span(self._flow.name) as span:
+        with otel_tracer.start_as_current_span(self._flow.name) as span, self._record_keyboard_interrupt_to_span(span):
```
zzn2, line 850: Do we do this only for "Flow" spans? How about a flex flow that does not have a "Flow" span?

huaiyan: Good catch. Maybe open another PR for flex scenario.
This PR is enough dangerous.

huaiyan: Another PR: https://github.com/microsoft/promptflow/pull/2918

src/promptflow-core/promptflow/executor/flow_executor.py
```diff
@@ -874,6 +874,16 @@ def _exec_inner_with_trace(
...
             return output, aggregation_inputs
 
+    @contextlib.contextmanager
+    def _record_keyboard_interrupt_to_span(self, span: Span):
+        try:
+            yield
+        except KeyboardInterrupt as ex:
+            if span.is_recording():
+                span.record_exception(ex)
+                span.set_status(StatusCode.ERROR, "KeyboardInterrupt received, execution cancelled.")
```
guming-learning, line 884: KeyboardInterrupt is our internal implementation for all cases except when user actually hit ctrl-c locally.
The message should just be "Excecution cancelled."

src/promptflow-core/promptflow/executor/_flow_nodes_scheduler.py
```diff
@@ -51,45 +81,56 @@ def execute(
...
+        if threading.current_thread() is threading.main_thread():
+            signal.signal(signal.SIGINT, signal_handler)
+            signal.signal(signal.SIGTERM, signal_handler)
+        else:
+            flow_logger.info(
+                "Current thread is not main thread, skip signal handler registration in AsyncNodesScheduler."
+            )
         parent_context = contextvars.copy_context()
-        with ThreadPoolExecutor(
+        # If we use with ThreadPoolExecutor, we can't use KeyboardInterrupt to exit waiting.
```
thy09, line 92: Add more comments to explain the logic here.

src/promptflow-core/promptflow/executor/_flow_nodes_scheduler.py
```diff
@@ -23,6 +27,32 @@
...
+    # Specifically, it does not exit non-daemon thread
+    # os._exit https://docs.python.org/3/library/os.html#os._exit
+    # Exit the process with status n, without calling cleanup handlers, flushing stdio buffers, etc.
+    # Specifically, it stops process without waiting for non-daemon thread.
+    os._exit(0)
+
+
+def signal_handler(sig, frame):
+    """
+    We have no way to cancel running tasks in ThreadPoolExecutor, so we raise KeyboardInterrupt to exit the process.
```
thy09, line 47: Add more detail about why
1 we have a separate thread here
2 raise an exception here

And explain what if we do not do this?

src/promptflow-core/promptflow/executor/_flow_nodes_scheduler.py
```diff
@@ -51,45 +81,56 @@ def execute(
...
+                for each_future in completed_futures:
+                    del self._future_to_node[each_future]
+                if timeout_task and timeout_task.done():
+                    raise LineExecutionTimeoutError(self._context._line_number, line_timeout_sec)
+                self._execute_nodes(executor)
+            event.set()
+            executor.shutdown()
+        except Exception as e:
+            # When meet exception, mark event to exit timeout task.
+            event.set()
```
guming-learning, line 122: Why not shutdown here?

huaiyan: If call shudown here, will wait for running threads finish.
Maybe add shudown(wait=False) is anyway better than current code?

src/promptflow-core/promptflow/executor/_flow_nodes_scheduler.py
huaiyan: Fixed~

thy09 (approved):

guming-learning (approved):

PR #2918, opened 2024-04-22
Add span cancel handling for flex flow and node execution

huaiyan:
# Description
Record exception and error status for each span.
We handle root span's tatus in this PR:https://github.com/microsoft/promptflow/pull/2889

image

### Testing Guidelines


thy09 (dismissed):

src/promptflow-tracing/promptflow/tracing/_trace.py
```diff
@@ -27,6 +28,16 @@
 IS_LEGACY_OPENAI = version("openai").startswith("0.")
 
 
+@contextlib.contextmanager
+def _record_keyboard_interrupt_to_span(span: Span):
+    try:
+        yield
+    except KeyboardInterrupt as ex:
+        span.record_exception(ex)
```
zzn2, line 36: We might be able to add `if span.is_recording():` here to align with the default error handler of the otel sdk.

Below are picked-up code from the otel sdk:

```python
@contextmanager
def use_span(
    span: Span,
    end_on_exit: bool = False,
    record_exception: bool = True,
    set_status_on_exception: bool = True,
) -> Iterator[Span]:
    """Takes a non-active span and activates it in the current context.

    Args:
        span: The span that should be activated in the current context.
        end_on_exit: Whether to end the span automatically when leaving the
            context manager scope.
        record_exception: Whether to record any exceptions raised within the
            context as error event on the span.
        set_status_on_exception: Only relevant if the returned span is used
            in a with/context manager. Defines whether the span status will
            be automatically set to ERROR when an uncaught exception is
            raised in the span with block. The span status won't be set by
            this mechanism if it was previously set manually.
    """
    try:
        token = context_api.attach(context_api.set_value(_SPAN_KEY, span))
        print(f"==== attached, token={token}")
        try:
            yield span
        finally:
            print(f"==== detaching, token={token}")
            context_api.detach(token)

    except Exception as exc:  # pylint: disable=broad-except
        if isinstance(span, Span) and span.is_recording():
            # Record the exception as an event
            if record_exception:
                span.record_exception(exc)

            # Set status in case exception was raised
            if set_status_on_exception:
                span.set_status(
                    Status(
                        status_code=StatusCode.ERROR,
                        description=f"{type(exc).__name__}: {exc}",
                    )
                )

        # This causes parent spans to set their status to ERROR and to record
        # an exception as an event if a child span raises an exception even if
        # such child span was started with both record_exception and
        # set_status_on_exception attributes set to False.
        raise

    finally:
        if end_on_exit:
            span.end()
```

liucheng-ms (approved):

src/promptflow-tracing/promptflow/tracing/_trace.py
huaiyan: Good catch! Fixed.

Adrian999999999 (approved):

zzn2 (approved):

liucheng-ms (approved):

PR #3074, opened 2024-04-30
Add util method to summarize trace telemetry.

huaiyan:
# Description
Add util method to summarize trace count telemetry.

For long term telemetry, we will need span count.
The simplest solution is create custom event for each trace id and include span count in custom dimension.
But that may generate too many customer events to affect all telemetry.
So, we only record trace count first, and decide how to add span count later according to trace count's telemetry.
Maybe just decide a reasonable sampling rate.

### Testing Guidelines


riddlexu (dismissed):

src/promptflow-devkit/promptflow/_sdk/_trace_telemetry.py
```diff
@@ -0,0 +1,36 @@
+import typing
```
zhengfeiwang, line 1: how about moving this to `/src/promptflow-devkit/promptflow/_sdk/_utilities/tracing_utils.py`? which I think is a better location to place util funcs

src/promptflow-devkit/promptflow/_sdk/_trace_telemetry.py
```diff
@@ -0,0 +1,36 @@
...
+            # We may need another field to indicate the language in the future, e.g. python, csharp.
+            execution_target = span.attributes.get(SpanAttributeFieldName.EXECUTION_TARGET, "code")
+
+            scenario = "script"
+            if SpanAttributeFieldName.BATCH_RUN_ID in span.attributes:
+                scenario = "batch"
+            elif SpanAttributeFieldName.LINE_RUN_ID in span.attributes:
+                scenario = "test"
+
+            key = (subscription_id, resource_group, workspace_name, scenario, execution_target)
```
zhengfeiwang, line 33: not sure is it possible to use `collections.namedtuple` here? raw tuple may lead to some unexpected usage

huaiyan: Good catch! Thanks!

zhengfeiwang (approved):

Hhhilulu (approved):

PR #3120, opened 2024-05-06
Enhance Traceability of Generator Outputs in Promptflow Tracing

liucheng-ms:
# Description

This PR introduces an enhanced method for capturing the trace of generator output functions within the Promptflow tracing framework. In the previous approach, when a function returned a generator, the associated span would terminate, and the output would be a string representing the generator object. Subsequently, a new span would be initiated when another function consumed this generator.

With the modifications proposed in this PR, the span persists beyond the point where a function returns a generator, concluding only after the generator has been fully consumed. Consequently, the output of the span encapsulating the generator or iterator object is now a comprehensive list of the iterated objects.

These changes significantly improve the transparency and intelligibility of the tracing process for generator outputs.

## Example with OpenAI Call

Consider the following code for a node:

```python
@trace
def chat(connection: AzureOpenAIConnection, question: str, stream: bool = False):
    connection_dict = normalize_connection_config(connection)
    client = AzureOpenAI(**connection_dict)

    messages = [{"role": "system", "content": "You are a helpful assistant."}, {"role": "user", "content": question}]
    response = client.chat.completions.create(model="gpt-35-turbo", messages=messages, stream=stream)

    if stream:
        def generator():
            for chunk in response:
                if chunk.choices:
                    yield chunk.choices[0].delta.content or ""
        return "".join(generator())
    return response.choices[0].message.content or ""

@tool
def my_python_tool(connection: AzureOpenAIConnection, question: str, stream: bool) -> str:
    return chat(connection, question, stream)
```

With the changes in this PR, specifying `stream=True` yields a generator object from the `client.chat.completions.create` call.

### Original Implementation

image

- In the original implementation, two spans were generated for `openai_chat`. The first span ended when the call finished, and the output was a string representing the generator object. The second span started when the generator was consumed and ended when the generator was fully consumed.

### New Implementation

image

- In the new implementation, only one span is generated for `openai_chat`. The span ends when the generator is fully consumed.

### Testing Guidelines


src/promptflow-tracing/promptflow/tracing/_trace.py
```diff
@@ -179,53 +169,35 @@ def enrich_span_with_llm_if_needed(span, original_span, inputs, generator_output
...
-        generator_proxy = GeneratorProxy(generator)
-        yield from generator_proxy
-        generator_output = generator_proxy.items
-        enrich_span_with_llm_if_needed(span, original_span, inputs, generator_output)
-        enrich_span_with_openai_tokens(span, TraceType(original_span.attributes["span_type"]))
-        enrich_span_with_output(span, serialize_attribute(generator_output))
-        span.set_status(StatusCode.OK)
+def traced_generator(span, inputs, generator, trace_type):
+    generator_proxy = GeneratorProxy(generator)
+    yield from generator_proxy
```
thy09, line 174: catch exceptions to make sure the span could be ended even there are any errors?

src/promptflow-tracing/promptflow/tracing/_trace.py
```diff
@@ -402,13 +376,19 @@ async def wrapped(*args, **kwargs):
...
-                span.set_status(StatusCode.OK)
-                output = Tracer.pop(output)
+                if isinstance(output, AsyncIterator):
+                    output = Tracer.pop(output)
+                    output = traced_async_generator(span, trace.inputs, output, trace_type)
+                else:
+                    enrich_span_with_trace_type(span, trace.inputs, output, trace_type)
+                    span.set_status(StatusCode.OK)
+                    output = Tracer.pop(output)
             except Exception as e:
```
thy09, line 386: Maybe we could have something like `should_end` signal to keep the logic more clean.

src/promptflow-tracing/promptflow/tracing/_trace.py
```diff
@@ -402,13 +376,19 @@ async def wrapped(*args, **kwargs):
...
                 Tracer.pop(None, e)
                 raise
         token_collector.collect_openai_tokens_for_parent_span(span)
+        if not isinstance(output, AsyncIterator):
```
thy09, line 390: We should end the span even an exception is raised?

src/promptflow-tracing/promptflow/tracing/_trace.py
```diff
@@ -455,7 +435,9 @@ def wrapped(*args, **kwargs):
         span_name = get_node_name_from_context(used_for_span_name=True) or trace.name
         # need to get everytime to ensure tracer is latest
         otel_tracer = otel_trace.get_tracer("promptflow")
-        with otel_tracer.start_as_current_span(span_name) as span, _record_cancellation_exceptions_to_span(span):
+        with otel_tracer.start_as_current_span(
```
thy09, line 438: Maybe we can have our own `start_as_current_span` to avoid duplicated if/end logic in multiple places?

src/promptflow-tracing/promptflow/tracing/_trace.py
liucheng-ms: Sure.

src/promptflow-tracing/promptflow/tracing/_trace.py
liucheng-ms: Instead of using this signal, I'd like to wrap it in a function handle different output type.

src/promptflow-tracing/promptflow/tracing/_trace.py
weedqian: Try ... finally?

liucheng-ms: We plan to leverage the `start_as_current_span` method, which already handle the exception handling in the context manger to achieve this goal. Instead, we will implement our own start_as_current_span context manager that wraps logic of span lifecycle management.

src/promptflow-tracing/promptflow/tracing/_trace.py
liucheng-ms: Sure. Will do

src/promptflow-tracing/promptflow/tracing/_openai_utils.py
```diff
@@ -211,9 +211,14 @@ def __init__(self, response, is_chat):
...
-        for item in self._response:
-            if hasattr(item, "model"):
-                return item.model
+        """
+        This method iterates over each item in the _response list.
+        If the item has a non-empty 'model' attribute, it returns the model.
+        If no such item is found, it returns None.
+        """
+        for response_item in self._response:
+            if hasattr(response_item, "model") and bool(response_item.model):
```
thy09, line 220: Why `bool()` here?

src/promptflow-tracing/promptflow/tracing/_span_wrapper.py
```diff
@@ -0,0 +1,71 @@
...
+        self._self_should_end = True
+
+    def __getattr__(self, name: str) -> object:
+        """
+        Get attribute from the wrapped object.
+
+        :param name: The name of the attribute.
+        :return: The attribute of the wrapped object.
+        """
+        if name == "__wrapped__":
```
thy09, line 27: Looks hard to understand this part, please add more clear explanation.

src/promptflow-tracing/promptflow/tracing/_openai_utils.py
liucheng-ms: There is empty model in response, which is `''`. This will lead model of streaming mode to be empty in UI.

src/promptflow-tracing/promptflow/tracing/_span_wrapper.py
liucheng-ms: Removed to use private attribute on `Span` itself instead.

src/promptflow-core/promptflow/executor/flow_executor.py
```diff
@@ -880,7 +885,7 @@ def _get_node_referenced_flow_inputs(
     @contextlib.contextmanager
     def _start_flow_span(self, inputs: Mapping[str, Any]):
         otel_tracer = otel_trace.get_tracer("promptflow")
-        with otel_tracer.start_as_current_span(self._flow.name) as span:
+        with start_as_current_span(otel_tracer, self._flow.name) as span:
```
zzn2, line 888: Also consider moving the line above `otel_tracer = otel_trace.get_tracer("promptflow")` into this newly created function?

src/promptflow-tracing/promptflow/tracing/_trace.py
```diff
@@ -41,6 +41,71 @@ def _record_cancellation_exceptions_to_span(span: Span):
...
+            context,
+            kind,
+            attributes,
+            links,
+            start_time,
+            record_exception,
+            set_status_on_exception,
+            end_on_exit=False,
+        ) as span:
+            object.__setattr__(span, "__should_end", True)
```
thy09, line 97: Why not `setattr(span, "___should_end")`?

liucheng-ms: Sure. It's better one.

src/promptflow-tracing/promptflow/tracing/_openai_utils.py
zzn2: How about use the str itself to check, not using bool?
```suggestion
            if hasattr(response_item, "model") and response_item.model:
```

src/promptflow-tracing/promptflow/tracing/_trace.py
```diff
@@ -41,6 +41,71 @@ def _record_cancellation_exceptions_to_span(span: Span):
...
+            yield span
+
+    except (KeyboardInterrupt, asyncio.CancelledError) as ex:
+        handle_span_exception(span, ex)
```
zzn2, line 101: Add comment here to say that these exceptions are not handled by the otel framework?

src/promptflow-tracing/promptflow/tracing/_trace.py
```diff
@@ -41,6 +41,71 @@ def _record_cancellation_exceptions_to_span(span: Span):
...
+        else:
+            return traced_async_generator(span, inputs, output, trace_type)
+    else:
+        enrich_span_with_trace_type(span, inputs, output, trace_type)
+        span.set_status(StatusCode.OK)
+        return Tracer.pop(output)
+
+
+@contextlib.contextmanager
+def start_as_current_span(
```
zzn2, line 73: Shall we separate this PR into two, one is the update of the structure of trace related code (but no business changes), another is for the streaming LLM span update?

src/promptflow-tracing/promptflow/tracing/_trace.py
liucheng-ms: OK

src/promptflow-core/promptflow/executor/flow_executor.py
liucheng-ms: will do this in next PR

src/promptflow-tracing/promptflow/tracing/_trace.py
liucheng-ms: Thank you for the suggestion. Given our upcoming release, it's crucial to maintain a clear focus on our changes. While separating the PR could enhance clarity, merging them in this PR ensures we streamline our workflow and avoid potential delays.

zzn2 (approved):

PR #3125, opened 2024-05-07
Change openai chat api parameters default value to None 

dorisjoy:
Change openai chat api parameters default value to None, so that when user doesn't set these values, we can ignore and not pass these parameters to do the api call.

### DIFF
--- a/src/promptflow-tools/promptflow/tools/openai.py
+++ b/src/promptflow-tools/promptflow/tools/openai.py
@@ -100,8 +100,8 @@ class OpenAI(ToolProvider):
         stream: bool = False,
         stop: list = None,
         max_tokens: int = None,
-        presence_penalty: float = 0,
-        frequency_penalty: float = 0,
+        presence_penalty: float = None,
+        frequency_penalty: float = None,
         logit_bias: dict = {},
         user: str = "",
         # function_call can be of type str or dict.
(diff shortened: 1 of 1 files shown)


src/promptflow-tools/promptflow/tools/openai.py
```diff
@@ -152,6 +150,10 @@ def chat(
             params["response_format"] = response_format
         if seed is not None:
             params["seed"] = seed
+        if presence_penalty:
```
chjinche, line 153: if presence_penalty is not None:

src/promptflow-tools/promptflow/tools/openai.py
```diff
@@ -123,8 +123,6 @@ def chat(
             "top_p": top_p,
             "n": n,
             "stream": stream,
-            "presence_penalty": presence_penalty,
-            "frequency_penalty": frequency_penalty,
             "user": user,
```
chjinche, line 126: maybe better to remove user and n?

dorisjoy: Discussed offline, removed n.

chjinche (approved):

chjinche (approved):

DaweiCai (approved):

PR #3239, opened 2024-05-13
Adding Support for Image Generation with AzureOpenAI 

marlenezw:
This PR adds support for image generation with the AzureOpenAI API and Dalle. 
This capability was added to AzureOpenai in a recent release and more information can be found [here](https://learn.microsoft.com/en-us/azure/ai-services/openai/dall-e-quickstart?tabs=dalle3%2Ccommand-line&pivots=programming-language-python). 

The API should use the following structure that will work with Prompty and Promptflow 

```
client = AzureOpenAI(
    api_version="2024-02-01",  
    api_key=os.environ["AZURE_OPENAI_API_KEY"],  
    azure_endpoint=os.environ['AZURE_OPENAI_ENDPOINT']
)

result = client.images.generate(
    model="dalle3", # the name of your DALL-E 3 deployment
    prompt="a close-up of a bear walking throughthe forest",
    n=1
)
```

### DIFF
--- a/src/promptflow-core/promptflow/core/_prompty_utils.py
+++ b/src/promptflow-core/promptflow/core/_prompty_utils.py
@@ -190,6 +190,8 @@ def get_open_ai_client_by_connection(connection, is_async=False):
 def send_request_to_llm(client, api, parameters):
     if api == "completion":
         result = client.completions.create(**parameters)
+    if api == "images":
+        result = client.images.generate(model=parameters['model'], prompt=parameters['messages'][0]['content'], n=1)
     else:
         result = client.chat.completions.create(**parameters)
     return result


src/promptflow-core/promptflow/core/_prompty_utils.py
```diff
@@ -190,6 +190,8 @@ def get_open_ai_client_by_connection(connection, is_async=False):
 def send_request_to_llm(client, api, parameters):
     if api == "completion":
         result = client.completions.create(**parameters)
+    if api == "images":
```
wangchao1230, line 193: pls add some e2e test on this

src/promptflow-core/promptflow/core/_prompty_utils.py
```diff
@@ -190,6 +190,8 @@ def get_open_ai_client_by_connection(connection, is_async=False):
...
+        result = client.images.generate(model=parameters['model'], prompt=parameters['messages'][0]['content'], n=1)
```
wangchao1230, line 194: why is it hard coded a n parameter here?  How about other parameters?

marlenezw: I haven't added any other parameters because I just wanted something that works for our build demo. These are the 3 main parameters for a quick start. I'll try and in more if I can.

src/promptflow-core/promptflow/core/_prompty_utils.py
marlenezw: Will do

PR #3282, opened 2024-05-15
Change how we do the column mapping

nick863:
# Description

Change the column mapping logic:
1. If user specified target function and some column have outputs. prefix, raise the exception
2. Do not rename target generated columns, so they will always be marked by "outputs." prefix.
3. Rely only on _apply_column_mapping while running CodeClient; remove mapping logic from it, instead filter inputs in case evaluator does not accept **kwargs.
4. Add additional unit test and add test file for previous commit.

### Testing Guidelines


src/promptflow-evals/promptflow/evals/evaluate/_evaluate.py
```diff
@@ -90,6 +93,8 @@ def _validate_columns(
     :paramtype target: Optional[Callable]
     """
     if target:
+        if any(c.startswith(_OUTPUTS) for c in df.columns):
```
singankit, line 96: What is ther reason for this restriction ?

nick863: traget will generate columns, which starts with "outputs." and if user will set this column, it will be deleted from the output, however some evaluator may rely on it and the result will be unexpected. Previously, we were mapping to the output from the run and that was not a problem.
It is an edge case, but everything is possible.

nick863: We may remove the limitation, when we will use pf.run.

src/promptflow-evals/promptflow/evals/evaluate/_code_client/code_client.py
```diff
@@ -30,12 +31,13 @@ class CodeClient:
     def __init__(self):
         self._thread_pool = ThreadPoolExecutor(thread_name_prefix="evaluators_thread")
 
-    def _calculate_metric(self, evaluator, input_df, column_mapping, evaluator_name):
+    def _calculate_metric(self, evaluator, input_df, evaluator_name):
```
ninghu, line 34: Nice. Seems like the column mapping will happen ahead of time, and we are no longer relying on pf.run for column mapping. Is my understanding correct?

src/promptflow-evals/promptflow/evals/evaluate/_evaluate.py
ninghu: Instead of fail the call, how about we give user a warning instead?

src/promptflow-evals/promptflow/evals/evaluate/_code_client/code_client.py
nick863: Yes, essentially it is correct, however we are doing it only for thread-pool execution and for validation purposes. But yes, we essentially mimic this functionality in pandas way.

src/promptflow-evals/promptflow/evals/evaluate/_evaluate.py
```diff
@@ -184,22 +179,21 @@ def _apply_column_mapping(source_df: pd.DataFrame, mapping_config: dict, inplace
...
-                    if col_outputs in source_df.columns:
-                        map_from_key = col_outputs
-
-                    # If column needs to be mapped to already existing column.
-                    if map_to_key in source_df.columns:
-                        columns_to_drop.add(map_to_key)
-                    column_mapping[map_from_key] = map_to_key
-        # If some columns, which has to be dropped actually can renamed,
-        # we will not drop it.
+                    # Target-generated columns alwauys starts from .outputs.
```
ninghu, line 184: typo

nick863: Fixed.

src/promptflow-evals/promptflow/evals/evaluate/_evaluate.py
nick863: I have marked target outputs as __outputs.

ninghu (approved):

PR #3380, opened 2024-06-06
[Perf] Evaluate API: Support parallelized evaluator batch run through pf.run

ninghu:
# Description

The change in this PR addresses the performance issues we were seeing with the Evaluate API. This is the first step in optimizing performance. The improvements include:

- Parallelizing the pf.run for evaluators. Previously, it ran sequentially, contributing to most of the latency.
- Addressing the slowness of the import Evaluate API, which was due to the import of MLClient.
- Using threads to infer signatures for eval batch runs instead of processes. (This change has been moved to a seperated PR: https://github.com/microsoft/promptflow/pull/3412)

**Improvements from this change:**
Windows OS, remote tracking disabled

- 4 evaluators, 100 rows:
    - Previous (pf.run without threadpool): 320 secs
    - Current (pf.run with threadpool): 78 secs (~75% improvement)

- 1 evaluator, 1 row:
   - Previous: 53 secs
   - Current: 17 secs (~68% improvement)

Investigation details can be found [here](https://microsoft-my.sharepoint.com/:w:/p/ninhu/ETB_zdMkFrdAuf3Lcg9ssrUB6RVmyuFs5Un1G74O1HlwSA?e=23ngPm)

# All Promptflow Contribution checklist:
- [ ] **The pull request does not introduce [breaking changes].**
- [ ] **CHANGELOG is updated for new features, bug fixes or other significant changes.**
- [ ] **I have read the [contribution guidelines](../CONTRIBUTING.md).**
- [ ] **Create an issue and link to the pull request to get dedicated review from promptflow team. Learn more: [suggested workflow](../CONTRIBUTING.md#suggested-workflow).**

## General Guidelines and Best Practices
- [ ] Title of the pull request is clear and informative.
- [ ] There are a small number of commits, each of which have an informative message. This means that previously merged commits do not appear in the history of the PR. For more information on cleaning up the commits in your PR, [see this page](https://github.com/Azure/azure-powershell/blob/master/documentation/development-docs/cleaning-up-commits.md).

### Testing Guidelines
- [ ] Pull request includes test coverage for the included changes.

PR #3412, opened 2024-06-13
[Perf] Evaluate API: Using threads to infer signatures for eval batch runs instead of processes. 

ninghu:
# Description

This change is related to https://github.com/microsoft/promptflow/pull/3380

### Testing Guidelines


huaiyan (dismissed):

src/promptflow-devkit/promptflow/_proxy/_python_inspector_proxy.py
```diff
@@ -35,7 +36,10 @@ def get_entry_meta(
         **kwargs,
     ) -> Dict[str, Any]:
         timeout = kwargs.get("timeout", FLOW_META_JSON_GEN_TIMEOUT)
-        load_in_subprocess = kwargs.get("load_in_subprocess", True)
+        load_in_subprocess = (
+            kwargs.get("load_in_subprocess", True)
+            and os.environ.get(PF_FLOW_META_LOAD_IN_SUBPROCESS, "True").lower() == "true"
```
huaiyan, line 41: It's quite not a common practice to check environment variable to determine a util method's behavior.
Maybe we could move this check operation to non-util method?
Not sure which code path you need to cover.
image

ninghu: The purpose of this environment variable is to offer a way to override the default flow meta loading behavior in the `get_entry_meta` through an environment variable. 

Since `get_entry_meta` is the initial function that introduces the `load_in_subprocess`, I think it’s an appropriate location for this override check. Moving this check to upper layer functions would require method signature changes in multiple places, also need to accommodate different code paths. Current approach tries to maintain code simplicity while fulfilling the needs of the change. Hope this makes sense.

src/promptflow-devkit/promptflow/_proxy/_python_inspector_proxy.py
```diff
@@ -35,7 +36,10 @@ def get_entry_meta(
         **kwargs,
     ) -> Dict[str, Any]:
         timeout = kwargs.get("timeout", FLOW_META_JSON_GEN_TIMEOUT)
-        load_in_subprocess = kwargs.get("load_in_subprocess", True)
+        load_in_subprocess = (
+            kwargs.get("load_in_subprocess", True)
```
wangchao1230, line 40: why do we need to have two ways to configure here? 
are both of them needed to be configured? is this by design?

ninghu: `load_in_subprocess` exists before but it is not being used anywhere at the moment. Removing it.

wangchao1230 (approved):

nick863 (approved):

src/promptflow-evals/promptflow/evals/evaluate/_batch_run_client/batch_run_context.py
```diff
@@ -23,10 +23,12 @@ def __enter__(self):
...
             os.environ[PF_FLOW_ENTRY_IN_TMP] = "true"
+            os.environ[PF_FLOW_META_LOAD_IN_SUBPROCESS] = "false"
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         if isinstance(self.client, CodeClient):
             recover_openai_api()
 
         if isinstance(self.client, ProxyClient):
             os.environ.pop(PF_FLOW_ENTRY_IN_TMP, None)
+            os.environ.pop(PF_FLOW_META_LOAD_IN_SUBPROCESS, None)
```
singankit, line 34: Should we restore it to it original value ?

PR #3529, opened 2024-07-09
Make evaluation run a context manager instead of a singleton.

nick863:
# Description

In this PR we are doing three things:
1. Making evaluator run a context manager
2. Moving function, writing properties to the run history to the evaluation run as a method
3. Removing code to keep evaluator run as a singletone as we, generally, do not need it.
4. Doing some unittest refactoring and adding/removing appropriate tests.
See work item 3334452.

### Testing Guidelines

### DIFF
--- a/src/promptflow-evals/promptflow/evals/evaluate/_eval_run.py
+++ b/src/promptflow-evals/promptflow/evals/evaluate/_eval_run.py
@@ -10,7 +10,7 @@ import posixpath
 import requests
 import time
 import uuid
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Set
 from urllib.parse import urlparse
 
 from requests.adapters import HTTPAdapter
@@ -142,8 +142,9 @@ class EvalRun(contextlib.AbstractContextManager):
         """
         Start the run, or, if it is not applicable (for example, if tracking is not enabled), mark it as started.
         """
-        if self._status != RunStatus.NOT_STARTED:
-            raise ValueError("The run has already started. Please end this run and to start another one.")
+        self._check_state_and_log('start run',
+                                  {v for v in RunStatus if v != RunStatus.NOT_STARTED},
+                                  True)
         self._status = RunStatus.STARTED
         if self._tracking_uri is None:
             LOGGER.warning("tracking_uri was not provided, "
@@ -196,7 +197,9 @@ class EvalRun(contextlib.AbstractContextManager):
         :type reason: str
         :raises: ValueError if the run is not in ("FINISHED", "FAILED", "KILLED")
         """
-        if not self._check_state_and_log('stop run'):
+        if not self._check_state_and_log('stop run',
+                                         {RunStatus.BROKEN, RunStatus.NOT_STARTED, RunStatus.TERMINATED},
+                                         False):
             return
         if self._is_promptflow_run:
             # This run is already finished, we just add artifacts/metrics to it.
@@ -206,9 +209,6 @@ class EvalRun(contextlib.AbstractContextManager):
             raise ValueError(
                 f"Incorrect terminal status {reason}. " 'Valid statuses are "FINISHED", "FAILED" and "KILLED".'
             )
-        if self._status == RunStatus.TERMINATED:
-            LOGGER.warning("Unable to stop run because it was already terminated.")
-            return
         url = f"https://{self._url_base}/mlflow/v2.0" f"{self._get_scope()}/api/2.0/mlflow/runs/update"
         body = {
             "run_uuid": self.info.run_id,
@@ -309,22 +309,30 @@ class EvalRun(contextlib.AbstractContextManager):
             f"{response.text=}."
         )
 
-    def _check_state_and_log(self, action : str) -> bool:
+    def _check_state_and_log(
+            self,
+            action: str,
+            bad_states: Set[RunStatus],
+            should_raise: bool) -> bool:
         """
         Check that the run is in the correct state and log worning if it is not.
 
         :param action: Action, which caused this check. For example if it is "log artifact",
                        the log message will start "Unable to log artifact."
         :type action: str
+        :param bad_states: The states, considered invalid for given action.
+        :type bad_states: set
+        :param should_raise: Should we raise an error if the bad state has been encountered?
+        :type should_raise: bool
+        :raises: RuntimeError if should_raise is True and invalid state was encountered.
         :return: boolean saying if run is in the correct state.
         """
-        if self._status == RunStatus.NOT_STARTED:
-            LOGGER.warning(
-                f"Unable to {action}. The run did not started. "
-                "Please start the run by calling start_run method.")
-            return False
-        if self._status == RunStatus.BROKEN:
-            LOGGER.warning(f"Unable to {action} because the run failed to start.")
+        if self._status in bad_states:
+            msg = f"Unable to {action} due to Run status={self._status}."
+            if should_raise:
+                raise RuntimeError(msg)
+            else:
+                LOGGER.warning(msg)
             return False
         return True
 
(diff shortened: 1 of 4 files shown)


src/promptflow-evals/promptflow/evals/evaluate/_eval_run.py
```diff
@@ -304,6 +320,17 @@ def _log_warning(self, failed_op: str, response: requests.Response) -> None:
...
         )
 
+    def _raise_not_started_nay_be(self) -> None:
+        """
+        Raise value error if the run was not started.
+
+        :raises: ValueError
+        """
+        if self._status == RunStatus.NOT_STARTED:
+            raise ValueError(
```
singankit, line 330: When is this error raised ? And is customer expected to do something to resolve it ?

src/promptflow-evals/promptflow/evals/evaluate/_utils.py
```diff
@@ -90,43 +74,42 @@ def _log_metrics_and_instance_results(
     # Adding line_number as index column this is needed by UI to form link to individual instance run
     instance_results["line_number"] = instance_results.index.values
 
-    ev_run = EvalRun(
+    with EvalRun(
```
ninghu, line 77: Maybe `EvalRunUploader` is better name?

src/promptflow-evals/promptflow/evals/evaluate/_eval_run.py
```diff
@@ -154,7 +137,31 @@ def _get_scope(self) -> str:
             self._workspace_name,
         )
 
-    def _start_run(self, run_name: Optional[str]) -> bool:
+    def start_run(self) -> None:
```
ninghu, line 140: If we don't expect the caller to use this method directly, change it to private.
Even better, we can get rid of start_end and end_run, move the implementation to enter and exit instead.

src/promptflow-evals/promptflow/evals/evaluate/_eval_run.py
```diff
@@ -314,7 +341,8 @@ def log_artifact(self, artifact_folder: str, artifact_name: str = EVALUATION_ART
         :param artifact_folder: The folder with artifacts to be uploaded.
         :type artifact_folder: str
         """
-        if self._is_broken:
+        self._raise_not_started_nay_be()
```
ninghu, line 344: Do we really need this check in all the log methods? Shouldn't we just print a warning saying something like, "Failed to create the run; run artifacts won't be uploaded. Please try again later."

nick863: This check is added for developers to avoid setting wrong status and mishandling of the run. Customer should not see this issue, as this is an internal class, however, if it happens our unit tests MUST fail.

src/promptflow-evals/promptflow/evals/evaluate/_eval_run.py
nick863: This check is added for the integrity, this class is only used internally and if this exception has raised, it means the bug. It is essentially done to fail the unit/e2e tests if that has somehow happened.

src/promptflow-evals/promptflow/evals/evaluate/_utils.py
nick863: EvalRun creates the dummy azure job under the hood, which is being operated as a run.

src/promptflow-evals/promptflow/evals/evaluate/_eval_run.py
nick863: Fixed.

src/promptflow-evals/promptflow/evals/evaluate/_utils.py
singankit: @nick863 This happens only when target is not present right ?

src/promptflow-evals/promptflow/evals/evaluate/_eval_run.py
ninghu: What's the point of showing a developer-facing message to the customer? This could cause confusion. If the run fails to be created in _start_run(), let the subsequent calls fail instead. Wouldn't that be better?

src/promptflow-evals/promptflow/evals/evaluate/_utils.py
nick863: Yes, otherwise, we just take the Promptflow run and use it for logging.

src/promptflow-evals/promptflow/evals/evaluate/_eval_run.py
nick863: In this case should be marked as broken, by our code the run should never be in Non-started state. This exception is added for code integrity, in case we will occasionally change this logic.

src/promptflow-evals/promptflow/evals/evaluate/_eval_run.py
nick863: In this case should be marked as broken, by our code the run should never be in Non-started state. This exception is added for code integrity, in case we will occasionally change this logic. If this issue will be only logged, we may not be able to catch it by the unit tests and customer will only not get the traces and will find out the error message in the logs. This error is primarily for us to avoid this state.

src/promptflow-evals/promptflow/evals/evaluate/_eval_run.py
nick863: Fixed, made it a warning.

src/promptflow-evals/promptflow/evals/evaluate/_eval_run.py
```diff
@@ -156,76 +138,97 @@ def _get_scope(self) -> str:
...
-        response = self.request_with_retry(
-            url=url,
-            method='POST',
-            json_dict=body
-        )
-        if response.status_code != 200:
-            self.info = RunInfo.generate(run_name)
-            LOGGER.warning(f"The run failed to start: {response.status_code}: {response.text}."
+        if self._status != RunStatus.NOT_STARTED:
+            raise ValueError("The run has already started. Please end this run and to start another one.")
```
ninghu, line 146: If the incorrect state is due to an inappropriate value being passed to a function, then ValueError makes sense. Otherwise, I think RuntimeError is probably a better fit.

And, can we reuse `_check_state_and_log` here as well?

src/promptflow-evals/promptflow/evals/evaluate/_eval_run.py
```diff
@@ -306,6 +309,25 @@ def _log_warning(self, failed_op: str, response: requests.Response) -> None:
             f"{response.text=}."
         )
 
+    def _check_state_and_log(self, action : str) -> bool:
```
ninghu, line 312: Should we fail the call instead of just logging a warning if the state is incorrect? 

I feel raising an exception here is better. And, customers will see the final error message. To avoid confusion, let's use a common message like "Unable to {action} due to Run status={status}" instead of saying "please do something...".

src/promptflow-evals/promptflow/evals/evaluate/_eval_run.py
nick863: Fixed.

src/promptflow-evals/promptflow/evals/evaluate/_eval_run.py
nick863: Fixed the message.
The run is not doing any further actions if incorrect state is reached and logs the corresponding error message according to previous suggestions. Should we still raise the error?

ninghu: When user run into incorrect state, it must be bug that developer should fix. Thus, I think raising the error is better than logging a warning message.

ninghu (approved):

singankit (approved):

PR #3546, opened 2024-07-13
Split installation tests from other e2e tests.

nick863:
# Description

In this PR we split the installation time tests from unit and e2e tests.
 - Unit tests and e2e tests are running now in editable install
 - Removed the building wheel step from the tests
 - Added new gate, which builds the wheel and then installs it into virtual environment and logging time.
 - Added test for logging, which, however, does not switched on in CI/CD.
 - Re-enable tests on python 3.9 in CI/CD as we do not depend on wheel in tests anymore i.e. we have separate test for creation of a functional wheel.
 
Example query to application insights:
```
let date_of_run=datetime("2024-07-02");
customEvents
| where timestamp > date_of_run
| where customDimensions has 'installation-test'
| extend parsed = parse_json(customDimensions)
| project 
      name,
      run_date=format_datetime(timestamp, 'yyyy-MM-dd'),
      metric = parsed['activity_name'],
      OS = strcat(parsed["OS"], "_", parsed["OS_release"]),
      python_version = tostring(parsed['python_version']),
      install_time_s=todecimal(parsed["install_time_s"])
| summarize performance = avg(install_time_s) by OS, name
| render columnchart with(
    kind=unstacked
)
```
Result:
image

See work item 3342618 for the inotial task and task 3339407 to enable tests on python 3.9. In this PR we are installing wheel with virtual env, which does not have issue mentioned in the bug. 

- [x] **I confirm that all new dependencies are compatible with the MIT license.**
- [X] **Create an issue and link to the pull request to get dedicated review from promptflow team. Learn more: [suggested workflow](../CONTRIBUTING.md#suggested-workflow).**

### Testing Guidelines


.github/workflows/promptflow-evals-e2e-test-azure.yml
```diff
@@ -14,21 +14,7 @@ env:
   WORKING_DIRECTORY: ${{ github.workspace }}/src/promptflow-evals
 
 jobs:
-  build:
```
ninghu, line 17: Don't we need the same change for other evals pipelines, like regression test and e2e-test-local?

.github/workflows/promptflow-evals-installation-test.yml
```diff
@@ -0,0 +1,64 @@
+name: promptflow-evals-installation-test
```
ninghu, line 1: Can we use this chance to rename regression-test to performance-test? Also, I sort of remember that we already capture the installation time in regression test. Can we remove the code there if that's duplicated with the change here?

.github/workflows/promptflow-evals-e2e-test-azure.yml
nick863: Yes, I have removed this code in all other tests, so now they all use editable install.

.github/workflows/promptflow-evals-installation-test.yml
nick863: Renamed and removed the installation -time code from performance test.

singankit (dismissed):

ninghu (approved):

wangchao1230 (approved):

PR #3627, opened 2024-08-07
new error message and test

MilesHolland:
Adds some extra error checking to when we try to load an evaluator to recognize when a loaded evaluator has relative imports (which fail post-load in a non-trivial-to-fix way due to how our save/load system works). When this error is recognized, the resulting user-facing message is a bit more detailed to help user's fix their evaluator.

Also adds a test to ensure that this works as expected.


nagkumar91 (approved):

src/promptflow-core/promptflow/_core/tool_meta_generator.py
```diff
@@ -261,10 +261,21 @@ def load_python_module_from_file(src_file: Path):
...
     except Exception as e:
+        message = "Failed to load python module from file '{src_file}': {error_type_and_message}"
+        if (
+            isinstance(e, ImportError)
+            and hasattr(e, "msg")
+            and e.msg == "attempted relative import with no known parent package"
+        ):
+            message = (
+                "Failed to load python module from file '{src_file}' due to relative import."
+                + " Relative imports fail in evaluators that are saved and loaded. Please use absolute imports instead."
```
ninghu, line 272: Can we move this logic to the upper layer? pf-core should not know the evaluators.

MilesHolland: Re-worded to not mention evaluators.

ninghu (approved):

src/promptflow-core/promptflow/_core/tool_meta_generator.py
```diff
@@ -261,10 +261,23 @@ def load_python_module_from_file(src_file: Path):
...
+        message = "Failed to load python module from file '{src_file}': {error_type_and_message}"
+        if (
+            isinstance(e, ImportError)
+            and hasattr(e, "msg")
+            and e.msg == "attempted relative import with no known parent package"  # pylint: disable=no-member
+        ):
+            message = (
+                "Failed to load python module from file '{src_file}' due to relative import."
+                + " Relative imports fail in loaded code that is not a known python module."
+                + " Please use absolute imports instead (ex: 'import .utils' instead of 'import utils'),"
```
singankit, line 273: Is this not relative import? Looks like either wording or example is incorrect.

MilesHolland: reworded to avoid that confusion.

diondrapeck (approved):

ninghu (approved):

luigiw (approved):

singankit (approved):

src/promptflow-core/promptflow/_core/tool_meta_generator.py
```diff
@@ -261,10 +261,22 @@ def load_python_module_from_file(src_file: Path):
...
+        message = "Failed to load python module from file '{src_file}': {error_type_and_message}"
+        if (
+            isinstance(e, ImportError)
+            and hasattr(e, "msg")
+            and e.msg == "attempted relative import with no known parent package"  # pylint: disable=no-member
+        ):
+            message = (
+                "Failed to load python module from file '{src_file}' due to relative import."
+                + " Relative imports fail in loaded code that is not a known python module."
+                + " Please use absolute imports."
```
zzn2, line 273: Is there other cases that raise `ImportError` (i.e. a misspelling in the package name) here, though it is an absolute import? Is this error message always accurate for all the cases?

zzn2 (approved):

src/promptflow-core/promptflow/_core/tool_meta_generator.py
MilesHolland: Good point, there could be other sources for import errors (like the user just having a bug in their own save/loaded eval. I'll change the wording to be speculative.

PR #3659, opened 2024-08-14
Move eval rai call to shared package

MilesHolland:
Moves the code formally located in _content_safety_common to promptflow.evals.common, since new evaluators will soon need them.


src/promptflow-evals/tests/evals/unittests/test_content_safety_rai_script.py
```diff
@@ -7,12 +7,8 @@
 import pytest
 from azure.identity import DefaultAzureCredential
 
-from promptflow.evals.evaluators._content_safety.common.constants import (
-    EvaluationMetrics,
-    HarmSeverityLevel,
-    RAIService,
-)
-from promptflow.evals.evaluators._content_safety.common.evaluate_with_rai_service import (
+from promptflow.evals.common.constants import EvaluationMetrics, HarmSeverityLevel, RAIService
```
ninghu, line 10: I think it's better to make it as an internal namespace: `promptflow.evals._common.constants`

src/promptflow-evals/promptflow/evals/evaluators/_content_safety/_hate_unfairness.py
```diff
@@ -2,13 +2,12 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # ---------------------------------------------------------
 from promptflow._utils.async_utils import async_run_allowing_running_loop
+from promptflow.evals.common.constants import EvaluationMetrics
 
 try:
-    from .common import ContentSafetyEvaluatorBase
-    from .common.constants import EvaluationMetrics
+    from ._content_safety_base import ContentSafetyEvaluatorBase
```
ninghu, line 8: Any downside if we also move content safety base to common?

MilesHolland: I considered that - my main reason against was that unlike the other files, which can all be extended to be used outside of a content safety content, the Content Safety Base is explicitly tied to the content safety evals, so having it remain among them seemed logically sound.

src/promptflow-evals/tests/evals/unittests/test_content_safety_rai_script.py
MilesHolland: done

ninghu (approved):

kdestin (approved):

PR #3702, opened 2024-08-30
changes to align with spec

MilesHolland:
A collection of small changes to the ECI and IP evaluators to match the spec more closely. Changes include:
- refactor non-content safety evaluator response parsing to add metric prefix to results. This includes special logic to capitalize the 'eci' prefix.
- Changed ECI simulator enum to just be 'ECI'.

### DIFF
--- a/src/promptflow-evals/promptflow/evals/_common/rai_service.py
+++ b/src/promptflow-evals/promptflow/evals/_common/rai_service.py
@@ -207,7 +207,7 @@ def parse_response(  # pylint: disable=too-many-branches,too-many-statements
     :return: The parsed annotation result.
     :rtype: List[List[Dict]]
     """
-
+    # non-numeric metrics
     if metric_name in {EvaluationMetrics.PROTECTED_MATERIAL, _InternalEvaluationMetrics.ECI}:
         if not batch_response or len(batch_response[0]) == 0 or metric_name not in batch_response[0]:
             return {}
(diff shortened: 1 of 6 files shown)


diondrapeck (dismissed):

src/promptflow-evals/promptflow/evals/evaluators/_eci/_eci.py
```diff
@@ -65,8 +65,8 @@ class ECIEvaluator:
     .. code-block:: python
 
         {
-            "label": "False",
-            "reasoning": "Some reason."
+            "ECI_score": "False",
```
luigiw, line 68: Hmmm, we call it a score, but the value is boolean?

MilesHolland: This is aligning with the spec, I'm neutral on changing this - on one hand. It's consistent with other evaluators. On the other hand, it's odd. Minsoo was also neutral on this.

src/promptflow-evals/promptflow/evals/evaluators/_eci/_eci.py
```diff
@@ -65,8 +65,8 @@ class ECIEvaluator:
...
+            "ECI_reasoning": "Some reason."
```
singankit, line 69: Should this be camel case ?

MilesHolland: I don't believe so, extant evaluators already return this results with snake-cased keys, this is just following that pattern.

src/promptflow-evals/promptflow/evals/evaluators/_eci/_eci.py
MilesHolland: Reverted back to label.

src/promptflow-evals/promptflow/evals/_common/rai_service.py
```diff
@@ -216,12 +216,27 @@ def parse_response(  # pylint: disable=too-many-branches,too-many-statements
         response = response.replace("true", "True")
         parsed_response = literal_eval(response)
         result = {}
-        result["label"] = parsed_response["label"] if "label" in parsed_response else np.nan
-        result["reasoning"] = parsed_response["reasoning"] if "reasoning" in parsed_response else ""
+        metric_prefix = _get_metric_prefix(metric_name)
+        # Use label instead of score since these are assumed to be boolean results.
+        result[metric_prefix + "_label"] = parsed_response["label"] if "label" in parsed_response else np.nan
```
ninghu, line 221: `np.nan` is float type. I wonder if we should just return an empty string.

MilesHolland: Done

ninghu (approved):

diondrapeck (approved):

PR #3890, opened 2024-12-20
fix key for FastapiContextDataProvider context data provider

prise6:
# Description

Hello,
with v2 fastapi + managed online endpoint with AML, the data collector can't write anything because in `FlowMonitor` input_data are requested with `input_data = self.context_data_provider.get_request_data()`.  This method looks for a key called `input_data` in `FastapiContextDataProvider`. 
But the value of this key is `None` because we don't set it correctly in `get_score_router()` (see changes is PR).

Maybe i'm wrong ?


prise6:
Still an issue to fix. This PR shows how to fix it.

prise6:
Is it possible to have a comment/look from a maintainer :) ?

prise6:
Still an issue. Take a look...

PR #4086, opened 2026-02-28
fix: replace fillna inplace with where() to avoid pandas 2.x FutureWa…

### LINKED ISSUE #2702: [BUG] FutureWarning from fillna call
**Describe the bug**

When I use the azure-ai-generative SDK's evaluate function(), it calls PromptFlow code which results in this warning in my logs:

/Users/pamelafox/ai-rag-chat-evaluator/.venv/lib/python3.11/site-packages/promptflow/_sdk/operations/_local_storage_operations.py:516: FutureWarning: Setting an item of incompatible dtype is deprecated and will raise an error in a future version of pandas. Value '(Failed)' has dtype incompatible with float64, please explicitly cast to a compatible dtype first.
  outputs.fillna(value="(Failed)", inplace=True)  # replace nan with explicit prompt

I am on promptflow 1.6 due to an issue with 1.7 and evaluate SDK, but I don't know if this warning is fixed yet, as the code seems to be similar in main.

**How To Reproduce the bug**

I experience it when I'm using https://github.com/Azure-Samples/ai-rag-chat-evaluator, which calls the evaluate function from azure-ai-generative, which uses promptflow behind the scenes.

**Expected behavior**

I expect no warnings.

Aftabbs:
`outputs.fillna(value="(Failed)", inplace=True)` raises FutureWarning in pandas 2.x because filling a float64 column with a string value triggers implicit dtype coercion.  The same warning appears again after casting to object dtype when pandas silently downcasts the result back.

Replace the call with `outputs.astype(object).where(outputs.notna(), other="(Failed)")`. Using `.where()` on an object-typed DataFrame avoids both the original coercion warning and the secondary silent-downcasting warning, and correctly marks all NaN cells (i.e. failed line-runs) with the "(Failed)" sentinel string.

Fixes #2702

### DIFF
--- a/src/promptflow-devkit/promptflow/_sdk/operations/_local_storage_operations.py
+++ b/src/promptflow-devkit/promptflow/_sdk/operations/_local_storage_operations.py
@@ -513,7 +513,7 @@ class LocalStorageOperations(AbstractBatchRunStorage):
             # if all line runs are failed, no need to fill
             if len(outputs) > 0:
                 outputs = self._outputs_padding(outputs, inputs[LINE_NUMBER].tolist())
-                outputs.fillna(value="(Failed)", inplace=True)  # replace nan with explicit prompt
+                outputs = outputs.astype(object).where(outputs.notna(), other="(Failed)")  # replace nan with explicit prompt
                 outputs = outputs.set_index(LINE_NUMBER)
         return inputs, outputs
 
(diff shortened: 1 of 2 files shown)


Aftabbs:
yes i agree

Aftabbs:

PR #4131, opened 2026-04-10
fix: restore local PFS authentication by adding WSGI loopback identity middleware

### LINKED ISSUE #4130: Local PFS locks legitimate callers out of `/listsecrets` and `/Telemetries/` after auth fix in #4090
## Bug Description

The March 13 auth hardening in PR #4090 changed `local_user_only()` to trust only `request.environ["REMOTE_USER"]`, but the local PFS startup path still launches plain Waitress (`waitress.serve(...)`) without any middleware that injects `REMOTE_USER` for local requests.

Under normal WSGI rules, client-supplied headers are exposed as `HTTP_*` keys, not `REMOTE_USER`, so local callers never satisfy the guard. The two endpoints still protected by `@local_user_only` — `/v1.0/Connections/<name>/listsecrets` and `/v1.0/Telemetries/` — now return 403 even for the same desktop user who started PFS.

This is not an intentional breaking change — there is no corresponding documentation update, CHANGELOG entry, or migration path. The in-tree tests still send `X-Remote-User` (which no longer maps to `REMOTE_USER`), confirming the regression is unintentional.

## Root Cause

`src/promptflow-devkit/promptflow/_sdk/_service/utils/utils.py` L60-66:

```python
user = request.environ.get("REMOTE_USER")
if user != getpass.getuser():
    abort(403)
```

`src/promptflow-devkit/promptflow/_cli/_pf/_service.py`:
```python
waitress.serve(app, host=service_host, port=port, threads=PF_SERVICE_WORKER_NUM)
# No middleware to inject REMOTE_USER for local requests
```

## Evidence

1. **Gate reads `REMOTE_USER`** — `utils.py` L60
2. **No injection path** — `waitress.serve()` is called without any auth middleware
3. **Tests still use old contract** — `tests/sdk_pfs_test/utils.py` L66, L129, L231 and `test_telemetry_apis.py` L15 still send `X-Remote-User`
4. **Historical context confirms this was the API contract**:
   - PR #1191 only hid `X-Remote-User` from Swagger docs, did not remove runtime support
   - PR #1750 (Jan 2024) added telemetry API using the same `X-Remote-User` pattern
5. **No follow-up** — No subsequent PR/issue addresses the broken local auth path
6. **`#4092` is unrelated** — that's a listing/spam issue, not this regression

## Steps to Reproduce

1. Start local PFS: `pf service start`
2. Try to list secrets: `GET /v1.0/Connections/my-conn/listsecrets`
3. Response: **403 Forbidden** (even though you are the local user who started PFS)
4. Same for `POST /v1.0/Telemetries/`

## Expected Behavior

The security fix in #4090 should have simultaneously provided a new trusted identity mechanism for local PFS callers. Local requests from the user who started PFS should still be authorized.

## Suggested Fix

Keep the anti-spoofing fix from #4090, but also add a trusted local auth mechanism that actually populates `REMOTE_USER` in the PFS hosting stack — for example, a WSGI middleware that sets `REMOTE_USER` to `getpass.getuser()` for connections originating from `127.0.0.1`/`::1`. Update the test helpers and swagger text to reflect the real contract.

YizukiAme:
Fixes #4130

## Problem

PR #4090 correctly removed the trust of client-supplied `X-Remote-User` headers to prevent identity spoofing. However, it did not provide an alternative mechanism for the local Prompt Flow Service (PFS) to identify the local user. Since `waitress` (the WSGI server used by PFS) never sets `REMOTE_USER` in the WSGI environ, all local calls to `@local_user_only` endpoints (`/Connections/{name}/listsecrets`, `/Telemetries/`) are now rejected with 403.

## Fix

Add a `LocalUserMiddleware` WSGI middleware that:

1. **Sets `REMOTE_USER = getpass.getuser()`** only when `REMOTE_ADDR` is `127.0.0.1` or `::1` (loopback)
2. **Strips client-supplied `HTTP_REMOTE_USER` / `HTTP_X_REMOTE_USER`** headers to prevent spoofing
3. **Leaves non-loopback connections without `REMOTE_USER`** — external connections are correctly rejected

The middleware is applied in `create_app()` via `app.wsgi_app = LocalUserMiddleware(app.wsgi_app)`.

The existing `local_user_only` decorator is unchanged — it continues to check only `request.environ["REMOTE_USER"]`, which is now properly populated by the server-side middleware rather than trusting any client header.

## Security Properties

- ✅ Loopback-only: only `127.0.0.1` and `::1` get identity injection
- ✅ Header stripping: client-supplied `Remote-User` / `X-Remote-User` are removed before reaching the app
- ✅ No new trust surface: the fix adds server-side identity, not client-side
- ✅ Backwards compatible: `@local_user_only` behavior is preserved exactly

## Tests

4 regression tests in `test_local_user_auth_middleware.py`:
- IPv4 loopback → REMOTE_USER injected
- IPv6 loopback → REMOTE_USER injected  
- Non-loopback → no REMOTE_USER (403 expected)
- Spoofed headers → stripped, real OS user used

_This fix was developed with AI assistance and reviewed by a human._

### DIFF
--- a/src/promptflow-devkit/promptflow/_cli/_pf/_service.py
+++ b/src/promptflow-devkit/promptflow/_cli/_pf/_service.py
@@ -23,6 +23,7 @@ from promptflow._sdk._constants import (
     PF_SERVICE_WORKER_NUM,
 )
 from promptflow._sdk._service.app import create_app
+from promptflow._sdk._service.utils.local_user_auth import LocalUserAuthMiddleware
 from promptflow._sdk._service.utils.utils import (
     check_pfs_service_status,
     dump_port_to_config,
@@ -47,10 +48,16 @@ logger = get_cli_sdk_logger()
 app = None
 
 
+def _create_app_with_local_user_auth():
+    flask_app, _ = create_app()
+    flask_app.wsgi_app = LocalUserAuthMiddleware(flask_app.wsgi_app)
+    return flask_app
+
+
 def get_app(environ, start_response):
     global app
     if app is None:
-        app, _ = create_app()
+        app = _create_app_with_local_user_auth()
     if os.environ.get(PF_SERVICE_DEBUG) == "true":
         app.logger.setLevel(logging.DEBUG)
     else:
@@ -255,7 +262,7 @@ def _prepare_app_for_foreground_service(port, force_start, service_host):
     port = validate_port(port, force_start, service_host)
     global app
     if app is None:
-        app, _ = create_app()
+        app = _create_app_with_local_user_auth()
     if os.environ.get(PF_SERVICE_DEBUG) == "true":
         app.logger.setLevel(logging.DEBUG)
     else:
--- a/src/promptflow-devkit/promptflow/_sdk/_service/utils/local_user_auth.py
+++ b/src/promptflow-devkit/promptflow/_sdk/_service/utils/local_user_auth.py
@@ -1,0 +1,19 @@
+# ---------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# ---------------------------------------------------------
+
+import getpass
+
+_LOOPBACK_REMOTE_ADDRS = {"127.0.0.1", "::1"}
+
+
+class LocalUserAuthMiddleware:
+    """Inject the current OS user for local PFS loopback requests."""
+
+    def __init__(self, wsgi_app):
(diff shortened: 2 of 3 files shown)


YizukiAme:

YizukiAme:
:-)

YizukiAme:
cmt

PR #4189, opened 2026-06-22
fix(devkit): avoid dtype warning for failed outputs

### LINKED ISSUE #2702: [BUG] FutureWarning from fillna call

fengjikui:
## Summary
- Fill missing failed output values after converting output columns to object dtype.
- Preserve the line-number column while avoiding pandas incompatible-dtype FutureWarning.
- Add a regression test for numeric output columns with padded failed rows.

## Root cause
When `_outputs_padding` adds missing output rows, numeric output columns contain `NaN`. Filling the whole dataframe with the string `"(Failed)"` attempts to write a string into a float column, which triggers pandas' incompatible-dtype FutureWarning and will become an error in a future pandas release.

Fixes #2702

## Validation
- `src/promptflow-evals/.venv/bin/python -m pytest src/promptflow-devkit/tests/sdk_cli_test/unittests/test_local_storage_operations.py -q` (2 passed)

### DIFF
--- a/src/promptflow-devkit/promptflow/_sdk/operations/_local_storage_operations.py
+++ b/src/promptflow-devkit/promptflow/_sdk/operations/_local_storage_operations.py
@@ -504,6 +504,13 @@ class LocalStorageOperations(AbstractBatchRunStorage):
         res = res.sort_values(by=LINE_NUMBER, ascending=True)
         return res
 
+    @staticmethod
+    def _fill_failed_outputs(df: "DataFrame") -> "DataFrame":
+        output_columns = [column for column in df.columns if column != LINE_NUMBER]
+        df = df.copy()
+        df[output_columns] = df[output_columns].astype(object)
+        return df.fillna(value="(Failed)")
+
     def load_inputs_and_outputs(self) -> Tuple["DataFrame", "DataFrame"]:
         if not self._sdk_inputs_path.is_file() or not self._sdk_output_path.is_file():
             inputs, outputs = self._collect_io_from_debug_info()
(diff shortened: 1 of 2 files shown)


fengjikui:
This is still active. I revalidated the exact PR head against current `main@3928a727` (the base has not moved):

- branch remains mergeable
- `test_local_storage_operations.py`: 2 passed, including the failed-output dtype-warning regression

GitHub reports `viewerCanReopen: false` for my account, so I cannot follow the bot's reopen/remove-label instruction directly. Could a maintainer please reopen this PR and remove `no-recent-activity`?

PR #4204, opened 2026-07-20
fix(parallel): allow '=' inside _parse_prefixed_args values

### LINKED ISSUE #4174: BUG: ValueError crash in _parse_prefixed_args when argument value contains '='
## What breaks

`_parse_prefixed_args()` in `src/promptflow-parallel/promptflow/parallel/_config/parser.py` (line 88) crashes with `ValueError: too many values to unpack (expected 2)` whenever an argument value contains the `=` character. This affects common real-world inputs: base64-encoded values (which use `=` as padding), URLs with query strings, and any JSON/key-value pairs embedded in an argument value.

**Minimal repro:**
```python
from promptflow.parallel._config.parser import _parse_prefixed_args

# Base64 "Hello World" ends with '=' padding — this crashes:
_parse_prefixed_args(["--pf_input_token=SGVsbG8gV29ybGQ="], "--pf_input_")
# ValueError: too many values to unpack (expected 2)
```

**Traceback:**
```
File "parser.py", line 88, in _parse_prefixed_args
    arg_name, arg_value = arg.split("=")
ValueError: too many values to unpack (expected 2)
```

## Root cause

`arg.split("=")` with no `maxsplit` produces more than 2 elements when the value contains `=`. The fix is `arg.split("=", 1)` so only the first `=` is treated as the name/value delimiter and the remainder of the string is preserved as the value.

Solaris-star:
## Summary
`_parse_prefixed_args()` crashed with:

```
ValueError: too many values to unpack (expected 2)
```

whenever an argument value contained `=` (base64 padding, URL query strings, embedded JSON/key-value pairs).

## Change
Use `str.split("=", 1)` so only the first `=` separates name and value.

## Test plan
- [x] Local unit checks for base64 and query-string values
- [x] Added `src/promptflow-parallel/tests/unittests/test_config_parser.py`

Fixes #4174.

### DIFF
--- a/src/promptflow-parallel/promptflow/parallel/_config/parser.py
+++ b/src/promptflow-parallel/promptflow/parallel/_config/parser.py
@@ -85,7 +85,9 @@ def _parse_prefixed_args(args: List[str], prefix: str) -> Dict[str, str]:
     for _, arg in enumerate(args):
         if arg.startswith(prefix):
             if "=" in arg:
-                arg_name, arg_value = arg.split("=")
+                # maxsplit=1 so values may contain '=' (base64 padding, query
+                # strings, embedded JSON) without ValueError (#4174).
+                arg_name, arg_value = arg.split("=", 1)
                 if len(arg_name) > len(prefix):
                     parsed[arg_name[len(prefix) :]] = arg_value
             elif pre_arg_name is None:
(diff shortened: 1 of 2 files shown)


Solaris-star:

PR #4205, opened 2026-07-31
fix(context_utils): serialise _change_working_dir with a threading.Lock

### LINKED ISSUE #3553: [BUG] Race condition with global state in process pool
**Describe the bug**
Prompt flow appears to modify the global working directory and workers in the line execution process pool can get spawned with different working directories.

**How To Reproduce the bug**
I created an example project to reproduce the bug here: https://github.com/bhanson-techempower/promptflow-concurrency-bug

Some runs will succeed and others will randomly fail with: 

> Flow path .../promptflow-concurrency-bug/sub_flow2/sub_flow3 does not exist.

Due to the worker for that particular node being spawned after the working directory has been changed.

With the example project the bug is reproduced for me almost every time.

In our production application we're seeing it about half the time when running a batch of 20 runs.

**Expected behavior**
The flow executes successfully every time because prompt flow does not share extra global state between processes.

**Running Information:**
 - Promptflow Package Version using `pf -v`:
 
```bash
{
  "promptflow": "1.12.0",
  "promptflow-core": "1.12.0",
  "promptflow-devkit": "1.12.0",
  "promptflow-tracing": "1.12.0"
}

Executable '.../promptflow-concurrency-bug/venv/bin/python'
Python (Darwin) 3.11.3 (main, May 25 2023, 12:42:30) [Clang 11.1.0 ]
```

 - Operating System: macOS Sonoma 14.5

**Additional context**

- [References of _change_working_dir](https://github.com/search?q=repo%3Amicrosoft%2Fpromptflow+_change_working_dir&type=code)
- [References of getcwd](https://github.com/search?q=repo%3Amicrosoft%2Fpromptflow%20getcwd&type=code)
- [This comment also stuck out as suspicious](https://github.com/microsoft/promptflow/blob/93c211dc64adc901c6be31df33c63c3782fdecbf/src/promptflow-core/promptflow/executor/_flow_nodes_scheduler.py#L172)

We've worked around the issue by modifying the way we invoke the sub flows:

```python
from pathlib import Path
flow_path = Path(__file__).parent / "sub_flow1"
flow = load_flow(flow_path)
```

AloysJehwin:
Fixes #3553

`_change_working_dir` uses `os.chdir()` which is process-global. When the line-execution process pool spawns workers while the parent thread is mid-chdir for a different flow, workers inherit the wrong directory and fail with "Flow path ... does not exist".

Added a module-level `threading.Lock` that serialises the getcwd/chdir/yield/chdir sequence. Race confirmed without the lock, eliminated with it. 4 tests added, all passing, pre-commit clean.

### DIFF
--- a/src/promptflow-core/promptflow/_utils/context_utils.py
+++ b/src/promptflow-core/promptflow/_utils/context_utils.py
@@ -7,19 +7,30 @@ please avoid using them in service code!!!"""
 import contextlib
 import os
 import sys
+import threading
+
+# os.chdir is process-global; serialise concurrent uses so that threads in the
+# line-execution process pool cannot observe each other's intermediate cwd state.
+# See https://github.com/microsoft/promptflow/issues/3553
+_working_dir_lock = threading.Lock()
 
 
 @contextlib.contextmanager
 def _change_working_dir(path, mkdir=True):
-    """Context manager for changing the current working directory"""
-    saved_path = os.getcwd()
+    """Context manager for changing the current working directory.
+
+    Uses a module-level lock to prevent concurrent threads from observing
+    each other's intermediate ``os.chdir`` state.
+    """
     if mkdir:
         os.makedirs(path, exist_ok=True)
-    os.chdir(str(path))
-    try:
-        yield
-    finally:
-        os.chdir(saved_path)
+    with _working_dir_lock:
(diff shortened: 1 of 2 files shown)


AloysJehwin:

PR #4221, opened 2026-09-08
ci: pin AliSajid/random-wait-action to a full commit SHA

holistis:
## Summary

One third-party GitHub Action in this repo is referenced by a mutable
**branch** ref instead of a tag or commit SHA:

```yaml
- name: Random Wait
  uses: AliSajid/random-wait-action@main
```

This appears in `scripts/readme/ghactions_driver/workflow_steps/step_azure_login.yml.jinja2`
and 5 more Jinja2 templates under `scripts/readme/ghactions_driver/workflow_templates/`,
and is generated into 84 checked-in workflow files under `.github/workflows/`
(the `samples_*` workflows).

## Why this is worth a separate look from #4214/#4207

I know GitHub Actions pinning is already an active effort here (#4214 open,
#4207/#4190 closed). I checked their diffs first and none of them touch this
reference — their tooling looks to have matched version-tag-style refs
(`@v1`, `@v2`, `@v4`, ...), and `@main` doesn't match that shape since it's
not a version tag at all.

That actually makes it the single riskiest pin in the whole workflow set:
a tag at least requires the upstream maintainer (or a compromised account)
to deliberately retarget it, but a branch head moves on *every* commit to
that branch, with zero action required on the attacker's side beyond
pushing to `main`. Several of the workflows that use this step run with
`id-token: write` and Azure OIDC federated-credential secrets
(`AZURE_CLIENT_ID`/`AZURE_TENANT_ID`/`AZURE_SUBSCRIPTION_ID`,
`AOAI_API_KEY_TEST`), so any step in that job — this one included — can
mint a token for `api://AzureADTokenExchange` for the duration of the run.

This is the same risk class described in the GitHub Actions security
hardening guide and referenced in #4214/#4190 (tj-actions/changed-files,
codfish/semantic-release-action).

## What changed

Pinned to the commit currently tagged `v2.12.0` on that action's repo,
with a version comment, matching the `<sha> # <tag>` convention already
used by #4214/#4207 for the other actions in this repo:

```yaml
uses: AliSajid/random-wait-action@933d3635c56777d240b67abe794f1a5f7bdfd063 # v2.12.0
```

Applied to:
- the 6 Jinja2 templates (the actual source of truth, generated via
  `python3 readme.py`)
- the 84 already-generated `.github/workflows/samples_*.yml` files, so the
  checked-in files stay consistent with what regeneration would produce

No other lines were touched — every changed file is a single-line diff
(90 files, 90 insertions, 90 deletions). Verified all 120 workflow YAML
files still parse after the change (`yaml.safe_load` over every file in
`.github/workflows/`).

## Is this safe to merge?

Yes. `933d3635c56777d240b67abe794f1a5f7bdfd063` is the exact commit the
`v2.12.0` tag and (at the time of writing) the `main` branch both point to,
confirmed via the GitHub API — no behavioral change. The action's inputs
(`minimum`, `maximum`) are unchanged in that version.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/samples_configuration.yml
+++ b/.github/workflows/samples_configuration.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Create Aoai Connection
         run: pf connection create -f ${{ github.workspace }}/examples/connections/azure_openai.yml --set api_key="${{ secrets.AOAI_API_KEY_TEST }}" api_base="${{ secrets.AOAI_API_ENDPOINT_TEST }}"
       - name: Random Wait
-        uses: AliSajid/random-wait-action@main
+        uses: AliSajid/random-wait-action@933d3635c56777d240b67abe794f1a5f7bdfd063 # v2.12.0
         with:
           minimum: 1
           maximum: 99
--- a/.github/workflows/samples_connections.yml
+++ b/.github/workflows/samples_connections.yml
@@ -68,7 +68,7 @@ jobs:
             sed -i -e "s/\${azure_open_ai_connection.api_key}/${{ secrets.AOAI_API_KEY_TEST }}/g" -e "s/\${azure_open_ai_connection.api_base}/$gpt_base/g" run.yml
           fi
       - name: Random Wait
-        uses: AliSajid/random-wait-action@main
+        uses: AliSajid/random-wait-action@933d3635c56777d240b67abe794f1a5f7bdfd063 # v2.12.0
         with:
           minimum: 0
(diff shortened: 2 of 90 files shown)


holistis:

holistis:
Still active on my end, this fix is ready to go. Happy to make any changes if needed. It closes a real risk since several of the affected workflows have Azure OIDC secrets in scope

